PR #8, opened 2016-11-15
version 0.1.0: use capstone-sys, add option api, add detail api, add syntax, etc.

m4b:
Implements a number of very useful features I've needed in the course of writing an application using it as a disassembler backend:

* turns into a "high-level" api providing more Rust like wrappers around the system bindings, and leaving the system bindings to `capstone-sys`
* switch to capstone-sys for system bindings (this has a number of very nice benefits)
* enable option api
* enable switching syntax
* half-enable detail api (register read, write, etc., is available, arch information is not)
* allow building capstone from source for exotic targets with `build_capstone` feature like wasm32
* add arm disassembly test
* firm up api
* add better error enums, more rusty style
* clean up various components
* distinguish between 0 instructions and error during disassembly
* many other changes

This will close one issue and allow implementing another by exposing the required capstone-sys bindings fully.


m4b:
There's a minor issue with `capstone-sys` not correctly copying the artifacts when built from source and when not using cmake, so may want to hold off until i bump capstone-sys to `0.2.0` to fix this (or not).

m4b:
I've forked a capstone3 crate because I haven't heard anything or had this PR acknowledged, so closing. If you feel like giving ownership to the capstone crate I'd be more than happy to take over if you're busy, otherwise thanks for the work setting up the foundations, was good to build on! :)

Issue #7, opened 2016-06-09
Access to cs_detail?


waywardmonkeys:
For something I'm working with, I'd like to be able to access `cs_detail` and get back info about registers read and written. (It might be useful to just wrap the actual `cs_detail` struct and make it easy to get access to all of the various details?)

This will require having a binding to set `cs_option` to generate detail.

tmfink:
I'm currently working on exposing some of detail interface in my branch:
https://github.com/tmfink/capstone-rs/tree/cs-detail

I do not plan supporting `cs_reg_read` and `cs_reg_write` yet because they require instruction ids. The architecture-specific capstone includes (such as [x86.h](https://github.com/aquynh/capstone/blob/8925ff9bde4faad8fbb5ae499e38ba29159252a4/include/x86.h)) are very large. We need to automatically parse them.

PR #10, opened 2017-04-18
Expose some of the cs_detail interface

tmfink:
This is a first take at resolving #7. I did not break the interface but expose some of the functionality in the cs_detail.

I did not add any functionality that requires architecture-specific header information (such as instruction or register ids) because the architecture-specific headers are so large and need to be automatically parsed first.

### DIFF
--- a/src/capstone.rs
+++ b/src/capstone.rs
@@ -158,7 +158,7 @@ impl Capstone {
 
     /// Returns the current error that could arise from enabling `CS_ERR_SKIPDATA`.
     fn get_skipdata_error(insn: &Insn) -> Option<CsErr> {
-        if insn.get_id() == 0 {
+        if insn.id() == 0 {
             Some(CsErr::CS_ERR_SKIPDATA)
         } else {
             None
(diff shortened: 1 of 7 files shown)


src/capstone.rs
```diff
@@ -67,23 +115,196 @@ impl Capstone {
...
+        };
+
+        Some(group_name)
+    }
+
+    /// Returns the current error from not enabling `CS_OPT_DETAIL`.
+    fn detail_required_error(&self) -> Option<CsErr> {
+        if *self.cs_option_state
+                .get(&CsOptType::CS_OPT_DETAIL)
+                .unwrap() == CsOptValueBool::CS_OPT_OFF as libc::size_t {
```
richo-stripe: Can this ever be `None`? I was a bit lazy when I originally wrote some of this, so there are panics in the library but it'd be nice to avoid introducing any more

tmfink: `cs_option_state` is a `HashMap` that maintains the current state of options (such as `CS_OPT_DETAIL`). Because `cs_option_state` is initialized with every possible `CsOptType` in `new()` and `cs_option_state` entries are only modified (never deleted), the `get()` should always return `Some()`.

PR #11, opened 2017-05-14
Windows build

kichristensen:


build.rs
```diff
@@ -0,0 +1,26 @@
+extern crate cmake;
```
richo: Where does this come from? I'm a little unfamiliar with the build machinery?

richo:
Thanks! I'm a tiny bit skeptical of this approach, it seems confusing to me that `cargo build` will go off and fetch submodules and stuff, but if that's a common idiom then I'm fine with merging it.

tmfink:
Since we now consume the `capstone-sys` crate, any Windows support should be added to `capstone-sys`. I'm actually currently working on adding Windows support.

See [capstone-sys issue #7](https://github.com/capstone-rust/capstone-sys/issues/7) for details.

Issue #12, opened 2017-08-14
setting cs_option


aep:
how would i set a cs_option? the api doesnt seem to be exposed

tathanhdinh:
I think it is implemented in [`capstone.rs`](https://github.com/richo/capstone-rs/blob/master/src/capstone.rs#L82), but the method is not set as public yet.

tmfink:
The `Capstone` struct now has public methods like `set_mode()`, `set_detail()`, and `set_extra_mode()`.

PR #16, opened 2017-09-29
API refactor WIP

tmfink:
Refactors the API, including:
* Partition `Mode` enum into:
    - `Mode`: referred to as "basic_mode" by [Capstone documentation](http://www.capstone-engine.org/lang_c.html)
        - Example: `Arm`/`Thumb`, `32`/`64`.
    - `ExtraMode`: "Extra Mode" in Capstone documentation
        - Example: Arm `V8`, Sparc `V9`
    - `Endian`: `big`/`little`
* Separate `new()` and `new_raw()` for `Capstone`
    - `new()` uses the builder pattern to ensure that only valid values are passed for mode, endian, extra_mode, etc.
        - `Capstone::new().x86().mode(x86::ArchMode::Mode64).syntax(x86::ArchSyntax::Intel).build()`
        - `Capstone::new().ppc().mode(ppc::ArchMode::Mode32).endian(Endian::Big).build()`
    - `new_raw()` has an API similar the original `new()`
        - `Capstone::new_raw(Arch::X86, Mode::Mode64, NO_EXTRA_MODE, None)`
        - `Capstone::new_raw(Arch::PPC, Mode::Default, NO_EXTRA_MODE, Some(Endian::Big))`
* Add setters to modify mode, syntax, etc.
    - Handles issue #12 

Private changes:
* Add tests for each architecture

Todo:
- [X] Use builder pattern with traits
- [x] ~~Run-time checks to ensure that valid values are passed to `new_raw()`, `set_mode()`, etc.~~ (save for future work)
- [x] ~~Expose more of the architecture-specific `detail` structure~~ (save for future work)
- [x] Put architecture modules under `arch::` module

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,7 @@ To produce a short demonstration. More complex demos welcome!
 - Binding Author(s):
     - m4b <m4b.github.io@gmail.com>
     - Richo Healey <richo@psych0tik.net>
+    - Travis Finkenauer <tmfinken@gmail.com>
 
 You may find a [full list of contributors on Github](https://github.com/capstone-rust/capstone-rs/graphs/contributors).
 
--- a/examples/demo.rs
+++ b/examples/demo.rs
@@ -1,11 +1,17 @@
 extern crate capstone;
 
+use capstone::prelude::*;
+
 const CODE: &'static [u8] = b"\x55\x48\x8b\x05\xb8\x13\x00\x00";
 
-fn example() -> capstone::CsResult<()> {
-    let mut cs = capstone::Capstone::new(capstone::Arch::X86, capstone::Mode::Mode64)?;
-    cs.set_detail(true).unwrap();
-    cs.att();
+fn example() -> CsResult<()> {
+    let cs = Capstone::new()
+        .x86()
+        .mode(arch::x86::ArchMode::Mode64)
+        .syntax(arch::x86::ArchSyntax::Att)
+        .detail(true)
+        .build()?;
+
     let insns = cs.disasm_all(CODE, 0x1000)?;
     println!("Got {} instructions", insns.len());
     for i in insns.iter() {
--- a/src/capstone.rs
+++ b/src/capstone.rs
@@ -1,63 +1,146 @@
 use libc;
-use std::collections::HashMap;
 use std::convert::From;
 use std::ffi::CStr;
+use std::marker::PhantomData;
 use std::mem;
 use error::*;
+use arch::CapstoneBuilder;
 use capstone_sys::*;
-use constants::{Arch, Mode, OptValue};
-use instruction::{Insn, Instructions, Detail};
-
+use constants::{Arch, CsModeRepr, Endian, ExtraMode, Mode, OptValue, Syntax};
+use instruction::{Detail, Insn, Instructions};
 
 /// An instance of the capstone disassembler
+#[derive(Debug)]
 pub struct Capstone {
-    csh: csh, // Opaque handle to cs_engine
-    cs_option_state: HashMap<cs_opt_type, libc::size_t>, // maintains state set with cs_option
-    _arch: Arch,
+    /// Opaque handle to cs_engine
+    csh: csh,
+
+    mode: usize,
+    endian: usize,
+    syntax: usize,
+    extra_mode: usize,
+    detail_enabled: bool,
+
+    /// We *must* set `mode`, `extra_mode`, and `endian` at once because `capstone`
+    /// handles them inside the arch-specific handler. We store the bitwise OR of these flags that
+    /// can be passed directly to `cs_option()`.
+    raw_mode: usize,
+
+    /// Architecture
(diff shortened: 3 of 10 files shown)


README.md
```diff
@@ -65,6 +65,7 @@ To produce a short demonstration. More complex demos welcome!
 - Binding Author(s):
     - m4b <m4b.github.io@gmail.com>
     - Richo Healey <richo@psych0tik.net>
+    - Travis Finkenauer <tmfinken@gmail.com>
```
m4b: Either make this alphabet or put yourself at the top :)

examples/demo.rs
```diff
@@ -1,11 +1,13 @@
...
 const CODE: &'static [u8] = b"\x55\x48\x8b\x05\xb8\x13\x00\x00";
 
-fn example() -> capstone::CsResult<()> {
-    let mut cs = capstone::Capstone::new(capstone::Arch::X86, capstone::Mode::Mode64)?;
-    cs.set_detail(true).unwrap();
-    cs.att();
+fn example() -> CsResult<()> {
+    let mut cs = Capstone::new().x86().mode(x86::ArchMode::Mode64).build()?;
+    cs.set_detail(true)?;
+    cs.set_syntax(Syntax::Att)?;
```
m4b: why isn't detail in the builder?

Similarly, syntax; for same reason if you set the mode to `x86::ArchMode::Mode64` on `.arm()` its a failure, i think syntax makes sense too, just from ergonomics perspective.

It's like I create a builder; then build some more (which can all fail)

src/capstone.rs
```diff
@@ -1,63 +1,146 @@
...
-    }
+        let endian = match endian {
+            Some(endian) => cs_mode::from(endian) as usize,
+            None => 0,
+        };
+        let extra_mode = Self::extra_mode_value(extra_mode);
+
+        let err = unsafe {
+            let combined_mode: usize = csmode as usize | endian | extra_mode;
+            let combined_mode: cs_mode = mem::transmute(combined_mode as CsModeRepr);
```
m4b: Is transmute  necessary?

src/capstone.rs
```diff
@@ -1,63 +1,146 @@
...
+
+            let mut cs = Capstone {
+                csh: handle,
+                syntax,
+                endian,
+                mode: csmode as usize,
+                extra_mode,
+                detail_enabled,
+                raw_mode,
+                _arch: arch,
```
m4b: why is this `_arch` and not just `arch`?

README.md
tmfink: I decided to keep it alphabetical (ignoring case)  =)

examples/demo.rs
tmfink: I suppose I can add `detail()` to the builder. I originally did not to keep the code simpler, but at this point it doesn't matter. Should not be much code to add.

You actually can call `.syntax()`--I just need change the example. I agree that the builder is nice because code will fail to compile if you pass a nonsensical value to the builder's methods. We still have the `new_raw()` method if you *really* don't want to use a builder.

src/capstone.rs
tmfink: Yes, because Rust does not let us create an enum with a non-existent discriminant. In order to mitigate the risk of transmuting to the wrong size/sign, I defined the `CsModeRepr` type. There's a test to ensure that the size is the same in `src/constants.rs`.

src/capstone.rs
tmfink: Actually, according to `git blame`, you introduced `_arch` in b3869291.
=)

m4b: :sob:

tmfink:
@m4b Can you review these minor changes?

If you approve, then I'll squash and merge.

src/lib.rs
```diff
@@ -48,23 +48,36 @@
...
+/// Contains items that you probably want to always import
+///
+/// For example:
+///
+/// ```
+/// use capstone::prelude::*;
+/// ```
+pub mod prelude {
+    pub use {Capstone, CsResult};
+    pub use arch::{self, BuildsCapstone, BuildsCapstoneEndian, BuildsCapstoneExtraMode,
```
m4b: I don't know how common it is to have a module imported via a prelude, but it makes examples look nicer, so yea

examples/demo.rs
```diff
@@ -1,11 +1,17 @@
...
-fn example() -> capstone::CsResult<()> {
-    let mut cs = capstone::Capstone::new(capstone::Arch::X86, capstone::Mode::Mode64)?;
-    cs.set_detail(true).unwrap();
-    cs.att();
+fn example() -> CsResult<()> {
+    let cs = Capstone::new()
+        .x86()
+        .mode(arch::x86::ArchMode::Mode64)
+        .syntax(arch::x86::ArchSyntax::Att)
+        .detail()
```
m4b: I almost like `detailed` more here, but its fine

PR #19, opened 2017-10-25
Update demo.rs on README.md

gunyarakun:
`demo.rs` on `README.md` is very old and cannot be compiled even in version `v0.1.0`.
I updated the demo.rs which uses builder pattern.


tmfink:
Thanks for the attention to detail!

PR #20, opened 2017-10-29
Api improve

tmfink:
These are mostly minor changes.

Once this pull request goes through, I would like to do a release.

### DIFF
--- a/src/instruction.rs
+++ b/src/instruction.rs
@@ -1,6 +1,5 @@
-extern crate libc;
-
 use std::ffi::CStr;
+use std::os::raw::c_uint;
 use std::ptr;
 use std::str;
 use std::fmt::{self, Debug, Display, Error, Formatter};
@@ -37,7 +36,7 @@ impl Instructions {
 impl Drop for Instructions {
     fn drop(&mut self) {
         unsafe {
-            cs_free(self.ptr, self.len as libc::size_t);
+            cs_free(self.ptr, self.len as usize);
         }
     }
 }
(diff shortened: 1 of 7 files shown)


src/instruction.rs
```diff
@@ -37,7 +36,7 @@ impl Instructions {
 impl Drop for Instructions {
     fn drop(&mut self) {
         unsafe {
-            cs_free(self.ptr, self.len as libc::size_t);
+            cs_free(self.ptr, self.len as usize);
```
m4b: why did this change?

tmfink: To eliminate the libc crate dependency

m4b: I see.  Is that wise / portable though?

tmfink: I think in this case, it's okay. We only used simple `libc` types that have well-defined counterparts in Rust (like `size_t`, `usize`, `u8`) or are already defined in `std::os::raw` (like `c_uint`).

If we needed to use more complicated C types (such as structs) or call `libc` functions, then I would agree that we should use the `libc` crate.

m4b: I'm not sure that `u8` is easily translateable from C, e.g., most infamously, the `char` type, but if we're just using `size_t` in FFI then I guess it's ok?

Also not sure what benefit removing the dep precisely is, since ostensibly, `capstone-sys` will have libc as a dep (and expose types from it), and hence transitively, so will this.

Unless `libc` was also removed as dep in `capstone-sys`?

tmfink: Capstone uses `uint8_t` (instead of char), so our use of `u8` is fine. Like you pointed out, it's a similar story for `size_t`/`usize`.

In the default configuration, `capstone-sys` has no dependencies, so normally `libc` is not in the dependency graph. 

Even when non-default features are used such that `libc` is a transitive dependency, it's nice to not have update the `libc` version in `Cargo.toml`.  =)

PR #43, opened 2018-08-15
simplify stdin.lock()

tathanhdinh:
IMHO, `stdin` may not need to be mutable and the second `stdin` can be replaced by the value expression `stdin.lock()`.


tmfink:
Thanks for the improvement!

PR #44, opened 2018-08-17
simplify unhexed_bytes

tathanhdinh:
IMHO, `unhexed_bytes` can be simplified a little bit, the [credit](https://codereview.stackexchange.com/questions/201698/convert-string-of-hex-into-vector-of-bytes) should be given to [@shepmaster](https://github.com/shepmaster).

### DIFF
--- a/examples/cstool.rs
+++ b/examples/cstool.rs
@@ -130,21 +130,19 @@ where
 }
 
 /// Select only hex bytes from input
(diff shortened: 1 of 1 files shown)


tmfink:
Thanks for the submission!
But, why is this simpler?

tmfink:
I'm closing this now due to inactivity, but feel free to open another PR.

PR #45, opened 2018-09-15
Update README: capstone-sys no longer has cc / cmake features.

waywardmonkeys:


tmfink:
Thanks for the contribution!

PR #46, opened 2018-09-15
Allow alloc_system on current Rust.

waywardmonkeys:
Allocators are enabled on stable in Rust 1.28 and later, so this
now works there.


tmfink:
Thanks for the update!

PR #55, opened 2019-01-13
Added set_skipdata method

mothran:
Added the new method set_skipdata() to the Capstone type.  Capstone option: CS_OPT_SKIPDATA

I also added a simple test with two x86 instructions, one decoding to '.byte'.

### DIFF
--- a/capstone-sys/Cargo.toml
+++ b/capstone-sys/Cargo.toml
@@ -11,7 +11,11 @@ license = "MIT"
 readme = "README.md"
 repository = "https://github.com/capstone-rust/capstone-rs/tree/master/capstone-sys"
 
-exclude = [ "capstone/docs/**", "capstone/xcode/**", "capstone/msvc/**" ]
+exclude = [
+    "/capstone/docs/**",
+    "/capstone/msvc/**",
+    "/capstone/xcode/**",
+]
 
 [badges]
 travis-ci = { repository = "capstone-rust/capstone-sys" }


tmfink:
Thanks for the PR! Sorry for the late response.

I like the idea, but I'm not comfortable merging the PR as-is:
* The only way for a user to check if bytes were returned is to check if the mnemonic is ".byte"
* A user could gets an instruction out, which would not make sense for a byte

Alternatives:
* Make the `Capstone::disasm()` family of functions return an enum that includes variants for instructions and bytes
    - This introduces complexity and run-time overhead having to check the variant for each instruction
* Add an `Instruction::is_skipped_byte() -> CsResult<bool>` method that allows users to check whether bytes were skipped without checking for ".byte"
    - A little bit janky, but does not negatively impact those who don't use skipdata.

Also, I'm not sure about the skipdata feature itself. You would use it on bytes where you don't know what is code or data. But, data bytes that happen to be a valid instruction could be mistakenly interpreted as code. With architectures like X86 that have variable length instruction, you could hit alignment issues where capstone could return an instruction that overlaps with code and data.

I'm open to suggestions.

tmfink:
@mothran if you could add some tests (or add to your existing test) that the id is *_INVALID for skipped bytes, then I'll merge this PR.

tmfink:
Looks like Criterion does not support Rust 1.24.1 anymore.

You can bump the minimum required version. See fab2a4a7e0f9ba6848be678028fd0a559231014e for reference.

mothran:
Sure thing, I am not familiar with Criterion, where can I find its minimum supported version currently?

tmfink:
I recommend you use rustup to install earlier touching to test before
committing again.

https://github.com/bheisler/criterion.rs/blob/master/README.md#compatibility-policy

tmfink:
> Using 1.24.1 I was able to run 'cargo test' and 'cargo bench' just fine. Is it possible there is something else I am missing to validate the error on my end?

Your `Cargo.lock` could be referencing an older version of Criterion. To replicate, try deleting your `Cargo.lock` the testing.

Another way to solve this problem could be updating `Cargo.toml` to using a "maximum" version constraint with `<=`. You should be able to say something like `>= 0.2.0, <= 0.2.7`. That way, we will not need to bump the minimum required Rust toolchain version.

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-requirements

Restricting versions could cause problems with dependency graphs normally, but Criterion is *only* a *dev-dependency*, so it should not cause dependency resolution errors in downstream users (since capstone-rs will be used as a normal dependency).

tmfink:
replaced with #65

PR #62, opened 2019-03-29
don't package PDFs or Xcode projects

froydnj:
These files aren't used by the crate itself.  For people vendoring the
crate, these files just take up unnecessary space.


tmfink:
Good catch. While you are already modifying this, could you also:
- Exclude `msvc/`
- Update the "repository" key to point to https://github.com/capstone-rust/capstone-rs/tree/master/capstone-sys

tmfink:
Squashed/merged 49fd17b9b3ed4d920a49166d82d9064289ab0a05 bd61ad8444fa771bff13ae0cf302107a447dd35e

PR #76, opened 2019-10-03
capstone: add support for cs_ac_type in x86

hghwng:
I don't think that generating `cs_ac_type` as an enum is a good option — it's a bitflag in fact. I had to write a ton of if-else as a workaround. Suggestions are appreciated!

### DIFF
--- a/capstone-rs/src/arch/x86.rs
+++ b/capstone-rs/src/arch/x86.rs
@@ -1,10 +1,12 @@
 //! Contains x86-specific types
 
 use core::convert::From;
+use core::convert::TryInto;
 use core::{cmp, fmt, slice};
 
 use capstone_sys::{
-    cs_x86_op__bindgen_ty_1, x86_op_mem, x86_op_type, cs_x86, cs_x86_op};
+    cs_ac_type, cs_x86, cs_x86_op, cs_x86_op__bindgen_ty_1, x86_op_mem, x86_op_type,
+};
 pub use capstone_sys::x86_insn_group as X86InsnGroup;
 pub use capstone_sys::x86_insn as X86Insn;
 pub use capstone_sys::x86_reg as X86Reg;
@@ -17,8 +19,7 @@ pub use capstone_sys::x86_avx_rm as X86AvxRm;
 
 pub use crate::arch::arch_builder::x86::*;
 use crate::arch::DetailsArchInsn;
-use crate::instruction::{RegId, RegIdInt};
-
+use crate::instruction::{RegAccessType, RegId, RegIdInt};
 
 /// Contains X86-specific details for an instruction
 pub struct X86InsnDetail<'a>(pub(crate) &'a cs_x86);
@@ -45,6 +46,10 @@ impl X86OperandType {
 pub struct X86Operand {
     /// Operand size
     pub size: u8,
+
+    /// How is this operand accessed? NOTE: this field is irrelevant if engine
+    /// is compiled in DIET mode.
+    pub access: Option<RegAccessType>,
 
     /// AVX broadcast
     pub avx_bcast: X86AvxBcast,
@@ -208,9 +213,10 @@ impl Default for X86Operand {
     fn default() -> Self {
         X86Operand {
             size: 0,
+            access: None,
             avx_bcast: X86AvxBcast::X86_AVX_BCAST_INVALID,
             avx_zero_opmask: false,
-            op_type: X86OperandType::Invalid
+            op_type: X86OperandType::Invalid,
         }
     }
 }
@@ -220,6 +226,7 @@ impl<'a> From<&'a cs_x86_op> for X86Operand {
         let op_type = X86OperandType::new(op.type_, op.__bindgen_anon_1);
         X86Operand {
             size: op.size,
+            access: cs_ac_type(op.access as _).try_into().ok(),
             avx_bcast: op.avx_bcast,
             avx_zero_opmask: op.avx_zero_opmask,
             op_type,
--- a/capstone-rs/src/instruction.rs
+++ b/capstone-rs/src/instruction.rs
@@ -1,3 +1,4 @@
+use core::convert::TryFrom;
 use core::fmt::{self, Debug, Display, Error, Formatter};
 use core::marker::PhantomData;
 use core::slice;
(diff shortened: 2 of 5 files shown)


hghwng:
Hmm, i forgot the test :sweat_smile: Now it's fixed!

capstone-rs/src/arch/x86.rs
```diff
@@ -218,8 +223,23 @@ impl Default for X86Operand {
...
+        } else if op.access == CS_AC_READ as _ {
+            RegAccessType::Read
+        } else if op.access == CS_AC_WRITE as _ {
+            RegAccessType::Write
+        } else if op.access == CS_AC_WRITE as u8 | CS_AC_READ as u8 {
+            RegAccessType::ReadWrite
+        } else {
+            unreachable!()
+        };
+
```
tmfink: Move this logic into an `impl From<cs_ac_type> for RegAccessType` in `instruction.rs`. That way, multiple architectures can use it.

capstone-rs/src/instruction.rs
```diff
@@ -36,6 +36,19 @@ pub type RegIdInt = u16;
 #[derive(Copy, Clone, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
 pub struct RegId(pub RegIdInt);
 
+/// Represents how the register is accessed.
+#[derive(Copy, Clone, Debug, Hash, PartialEq, Eq)]
+pub enum RegAccessType {
```
tmfink: Can an access be not read nor write, but still valid?

I also wonder if we should eliminate the `Invalid` variant and return `Option<RegAccessType>`, which may make it less likely for users to ignore the "invalid" case.

tmfink: Add an `impl` block that includes getter methods `is_read()` and `is_write()` so that users don't have to write an in themselves.

capstone-rs/src/arch/x86.rs
```diff
@@ -218,8 +223,23 @@ impl Default for X86Operand {
 impl<'a> From<&'a cs_x86_op> for X86Operand {
     fn from(op: &cs_x86_op) -> X86Operand {
         let op_type = X86OperandType::new(op.type_, op.__bindgen_anon_1);
+
```
tmfink: To eliminate the possibility for undefined behavior, do a separate bitwise check for read then write. Then check for cases ReadWrite, Read, and Write. In the else block, return Invalid.

Otherwise, other higher order bits being set could result in undefined behavior.

capstone-rs/src/arch/x86.rs
```diff
@@ -4,7 +4,7 @@ use core::convert::From;
 use core::{cmp, fmt, slice};
 
 use capstone_sys::{
-    cs_x86_op__bindgen_ty_1, x86_op_mem, x86_op_type, cs_x86, cs_x86_op};
+    cs_ac_type, cs_x86_op__bindgen_ty_1, x86_op_mem, x86_op_type, cs_x86, cs_x86_op};
```
tmfink: We may want to modify `capstone-sys` to treat `cs_ac_type` as a bitflag (which bindgen supports).

tmfink:
Thanks for the PR! I just posted some review comments. Ask me if you have any questions.

tmfink:
To update capstone-sys pre-generated bindings:
- In `capstone-sys/build.rs`, add `cs_ac_type` to the `bitfield_enum` pattern
- See `capstone-sys/pre_generated/update-bindings.md` to update pre-generated bindings
- Please keep any pre-generated bindings update in capstone-sys change as a separate commit.

hghwng:
Thanks for the thoughtful comments — I learned a lot! I'll work on it soon.

capstone-rs/src/instruction.rs
hghwng: Actually, I implemented such getters at first, but deleted the `impl` before committing. I think the name `is_read` or `is_write` is misleading: If the instruction both reads and writes to the very same register, should `is_read()` return `true`? Changing the name to `contains_read` or `is_read_only` tells them apart, but the name reads strange.

tmfink: I think if the access is read and write, then both getters should return true. Make sure that the documentation comment is clear.

Maybe a better name is something like is_readable.

capstone-rs/src/arch/x86.rs
hghwng: Perhaps bitwise check is redundant. We only need to update it if capstone itself gets updated.

Currently, [`cs_ac_type`](https://github.com/aquynh/capstone/blob/fa17ed2453bd87f4aec72842571aed893c3a1963/include/capstone/capstone.h#L203) only has read and write operation defined. As the documentation clearly states that the operations are platform-independent, it's very unlikely that new bits will be added in the future.

Suppose that future-proof is *really* needed, the proposed API (`enum {R, W, RW}`) should be abandoned as a side-effect: we can't enumerate all combinations. Exporting `cs_ac_type` as a bitflag is future proof, but the downside is obvious: we are delegating the responsibility of decoding cs_ac_type to the user. This design spoils the consistent style of high-level API which `capstone-rs` exposes.

hghwng:
How about this?

capstone-rs/src/instruction.rs
```diff
@@ -36,6 +37,64 @@ pub type RegIdInt = u16;
...
+            return Err(());
+        }
+
+        let is_readable = (access & CS_AC_READ).0 != 0;
+        let is_writable = (access & CS_AC_WRITE).0 != 0;
+        Ok(match (is_readable, is_writable) {
+            (true, false) => RegAccessType::Read,
+            (false, true) => RegAccessType::Write,
+            (true, true) => RegAccessType::ReadWrite,
+            _ => unreachable!(),
```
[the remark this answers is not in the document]
tmfink: Please add a test where you manually construct an invalid `cs_ac_type` (like `cs_ac_type(1337)`) and verify that trying to create a `RegAccessType` gives an `Err()`.

capstone-rs/src/instruction.rs
```diff
@@ -36,6 +37,64 @@ pub type RegIdInt = u16;
...
+    }
+}
+
+impl TryFrom<cs_ac_type> for RegAccessType {
+    type Error = ();
+
+    fn try_from(access: cs_ac_type) -> Result<Self, Self::Error> {
+        if access == CS_AC_INVALID {
+            return Err(());
+        }
```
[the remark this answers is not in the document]
hghwng: IMHO, checking it first matches the semantics better, because the API of capstone only specifies that `CS_AC_INVALID` means "invalid". It's possible that an access is neither readable nor writable, semantically. (Although this scenario is not possible with current `capstone.h`)

tmfink:
@hghwng any update?

hghwng:
@tmfink Sorry about the long delay. On the other hand, the delay helps me to reread and rethink your thoughtful comments. Now all of them makes sense :+1:

capstone-rs/src/instruction.rs
```diff
@@ -363,3 +416,9 @@ impl<'a> Display for Instructions<'a> {
         Ok(())
     }
 }
+
+#[cfg(test)]
```
tmfink: In keeping with convention, could you put this in `mod test{...}`?

capstone-rs/src/instruction.rs
```diff
@@ -36,6 +37,58 @@ pub type RegIdInt = u16;
...
+        self == RegAccessType::WriteOnly || self == RegAccessType::ReadWrite
+    }
+}
+
+impl TryFrom<cs_ac_type> for RegAccessType {
+    type Error = ();
+
+    fn try_from(access: cs_ac_type) -> Result<Self, Self::Error> {
+        // Check for flags other than CS_AC_READ or CS_AC_WRITE.
+        let unknown_flag_mask = !(CS_AC_READ | CS_AC_WRITE).0;
```
tmfink: Get rid of this unknown_flag_mask check; the match statement default case handles it

hghwng: Without the check, `RegAccess::try_from(1337)` returns `Some(_)`, which fails the unit test you requested to add.

tmfink: You're right; my mistake.

capstone-rs/src/instruction.rs
tmfink: Great work so far! Fix this and I'll merge.

tmfink:
@hghwng great work! Thanks again.

PR #77, opened 2019-10-08
capstone: add cs_regs_access

### LINKED ISSUE #63: Support v4 cs_regs_access API
cs_regs_access API: https://www.capstone-engine.org/op_access.html

SWW13:
Adds support for `cs_regs_access`, fixes #63.

I'm not sure how to implement the tests yet, the `instructions_*` tests are a bit complicated on first sight.

### DIFF
--- a/capstone-rs/src/capstone.rs
+++ b/capstone-rs/src/capstone.rs
@@ -388,15 +388,7 @@ impl Capstone {
     /// 2. Skipdata is disabled
     /// 3. Capstone was not compiled in diet mode
     pub fn insn_regs_access<'s, 'i: 's>(&'s self, insn: &'i Insn) -> CsResult<InsnRegsAccess> {
-        if !self.detail_enabled {
-            Err(Error::DetailOff)
-        } else if insn.id().0 == 0 {
-            Err(Error::IrrelevantDataInSkipData)
-        } else if Self::is_diet() {
-            Err(Error::IrrelevantDataInDiet)
-        } else {
-            Ok(unsafe { insn.regs_access(self.csh()) })
-        }
+        insn.regs_access(self.csh())
     }
 
     /// Returns a tuple (major, minor) indicating the version of the capstone C library.
--- a/capstone-rs/src/capstone.rs
+++ b/capstone-rs/src/capstone.rs
@@ -11,7 +11,9 @@ use capstone_sys::*;
 use crate::constants::{Arch, Endian, ExtraMode, Mode, OptValue, Syntax};
 use crate::error::*;
 use crate::ffi::str_from_cstr_ptr;
-use crate::instruction::{Insn, InsnDetail, InsnGroupId, InsnId, Instructions, RegId};
+use crate::instruction::{
+    Insn, InsnDetail, InsnGroupId, InsnId, InsnRegsAccess, Instructions, RegId,
+};
 
 
 /// An instance of the capstone disassembler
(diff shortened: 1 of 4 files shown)


capstone-rs/src/instruction.rs
```diff
@@ -328,6 +345,64 @@ impl<'a> InsnDetail<'a> {
     }
 }
 
+impl<'a> InsnRegsAccess {
+    fn new(cs: csh, ins: &cs_insn) -> Self {
```
tmfink: Any function that can cause undefined behavior should be marked unsafe.

However, once you move the error checking (checking return value), then this shouldn't need to be marked unsafe.

capstone-rs/src/capstone.rs
```diff
@@ -378,6 +380,25 @@ impl Capstone {
...
     }
 
+    /// Returns `RegsAccess` structure for a given instruction
+    ///
+    /// Requires:
+    ///
+    /// 1. Instruction was created with detail enabled
+    /// 2. Skipdata is disabled
+    /// 3. Capstone was not compiled in diet mode
+    pub fn insn_regs_access<'s, 'i: 's>(&'s self, insn: &'i Insn) -> CsResult<InsnRegsAccess> {
```
tmfink: Instead of trying to do all of the error handling here, check the returned integer and convert that to a `CsResult`.
You can move that logic into 
That way, we don't have to manually track all possible errors that capstone should be handling anyway.

For examples, see `disasm()`. Also, `From<capstone_sys::cs_err::Type>` is implemented for `Error`, so you can use Error::from(capstone_sys_rc)` to do the error parsing for you.

capstone-rs/src/instruction.rs
```diff
@@ -328,6 +345,64 @@ impl<'a> InsnDetail<'a> {
...
+        let mut regs_read = [0u16; 64];
+        let mut regs_read_count = 0u8;
+        let mut regs_write = [0u16; 64];
+        let mut regs_write_count = 0u8;
+
+        unsafe {
+            cs_regs_access(
```
tmfink: check this return value to create a CsResult

capstone-rs/src/instruction.rs
tmfink: Have this return a `CsResult<Self>` instead

capstone-rs/src/instruction.rs
```diff
@@ -145,6 +145,14 @@ pub struct Insn<'a> {
 /// `ArchDetail` enum.
 pub struct InsnDetail<'a>(pub(crate) &'a cs_detail, pub(crate) Arch);
 
+/// Contains information about registers accessed by an instruction, either explicitly or implicitly
+pub struct InsnRegsAccess {
```
tmfink: Add `#[derive(Debug, Clone, PartialEq, Eq)]`

capstone-rs/src/instruction.rs
```diff
@@ -328,6 +345,64 @@ impl<'a> InsnDetail<'a> {
...
+        RegsIter(self.regs_write[..self.regs_write_count as usize].iter())
+    }
+
+    /// Returns the number of explicit and implicit write registers
+    pub fn regs_write_count(&self) -> u8 {
+        self.regs_write_count
+    }
+}
+
+impl Debug for InsnRegsAccess {
```
tmfink: You can just derive the Debug impl; see above comment

capstone-rs/src/instruction.rs
```diff
@@ -328,6 +345,64 @@ impl<'a> InsnDetail<'a> {
     }
 }
 
+impl<'a> InsnRegsAccess {
+    fn new(cs: csh, ins: &cs_insn) -> Self {
+        let mut regs_read = [0u16; 64];
```
tmfink: Use `cs_regs` type directly instead of manually declaring an array.

capstone-rs/src/instruction.rs
tmfink: If you can't derive `Debug` because the arrays are too big (many of these impls are only defined for arrays up to a certain size), then add a comment explaining why you can't derive `Debug`.

tmfink:
@SWW13 Thanks for the pull-request! This is great work.

If you have any questions related to review comments or capstone, feel free to reach out.

capstone-rs/src/instruction.rs
SWW13: `Debug, PartialEq, Eq` can't be derived because of the array with size `>32`, I added a comment

capstone-rs/src/instruction.rs
SWW13: Can you give an example how to initialize a variable with a type?

capstone-rs/src/instruction.rs
tmfink: Could you manually impl `PartialEq`, `Eq`?

capstone-rs/src/instruction.rs
tmfink: It looks like `cs_regs` is a type alias:

~~~rust
pub type cs_regs = [u16; 64usize];
~~~

Just add an explicit type instead of using type inference:

~~~rust
let mut regs_read: cs_regs = [0u16; 64];
~~~

tmfink:
Addresses #84

tmfink:
Thanks for the PR, but I decided to go with #149 as a starting point.

Issue #56, opened 2019-01-14
capstone ints and enums


mothran:
I ran into a problem with matching instruction id's to their enum counter parts defined in capstone-sys.  Because rust does not have a provide a simple/safe way to cast a u32 to an enum value matching something like instruction id's to a list of possible instructions seems to get tricky.

This applies to instruction ID's as well as reg_ids, op_types and all the other capstone-sys enums.

For example it would be nice to have code as clean as:

```rust
fn is_valid_ret(insn: &Insn) -> bool {
    match insn.id().0 {
        x86_insn::X86_INS_RET => true,
        _ => false
    }
}
```

Maybe the bindgen could be changed for capstone-sys to output constants instead of enums for these types of values?  Or am I missing something defined in capstone-sys?

mothran:
I was curious about what the process would be like so for consideration:
https://github.com/mothran/capstone-sys/commit/daf261f6b2b64ec025a886288d8a7ce88fe33ef5
and
https://github.com/mothran/capstone-rs/commit/55e67d8c713c58a22b32c6aa0740c9d2184d47f6

I did not update the documentation to clearly reflect the changes, and there are still a number of arch specific enums that would need to be converted to work cleanly in the model above.  It was more a PoC to show myself if it would be a fair way to tackle the above issue.

tmfink:
Thanks for filing the issue!

I see what you are saying, `Insn::id()` returns an `InsnId`, which just contains an integer value. However, the architecture-specific instruction enums like `X86Insn` cannot be created from integers, which makes matching more tedious.

Possible solutions:

1. **More constants**: (Your recommendation) change the arch instructions enums to constified modules
    - This reduces "type safety" since nothing would prevent you from matching Arm instruction ids values to X86 instructions.
2. **More enums**: change the `Insn` type to be an enum over architecture types. For example, there would be a `X86(X86Insn)` variant.
    - This would incur some runtime overhead to "unwrap" a `Insn` to a known variant such as `X86Insn`
3. **More enums & generic `Capstone`**: to avoid runtime overhead as with (2), we could change the `Capstone` struct to be generic over architecture types.
    - For example, we would know at compile-time that a `Capstone<X86>::disasm()` generates `X86Insn`.
4. **`X86Insn::from_insn_id()`**: Add a function to `X86Insn` (and other archs) to create an `X86Insn` from an `InsnId`.
    - We would need to be able to iterate over `capstone_sys::x86_insn` at compile-time.
[...]

PR #79, opened 2019-12-04
impl From<u32> for instructions enums

losynix:
Addresses issue https://github.com/capstone-rust/capstone-rs/issues/56 with your [solution 4](https://github.com/capstone-rust/capstone-rs/issues/56#issuecomment-455254742): Add a function to X86Insn (and other archs) to create an X86Insn from an InsnId.

Now mothran's [example](https://github.com/capstone-rust/capstone-rs/issues/56#issue-398715105) can be written like this:

```rust
fn is_valid_ret(insn: &Insn) -> bool {
    match X86Insn::from_insn_id(insn.id().0) {
        X86Insn::X86_INS_RET => true,
        _ => false
    }
}
```

Parsing and appending code to the generated bindings is a bit dirty, maybe Rust will have a good and safe way to cast integers to enums one day.


capstone-sys/build.rs
```diff
@@ -148,6 +152,62 @@ fn env_var(var: &str) -> String {
...
+    for cap in re.captures_iter(&bindings) {
+        impl_arch_enum.push_str(
+            &format!("impl {}_insn {{\n
+            pub fn from_insn_id(id: u32) -> {}_insn {{\n
+            match id {{\n",
+            &cap["arch"], &cap["arch"]));
+
+        // find instructions and their id
+        let re2 = Regex::new(
+            &format!("{}_INS_(?P<ins>[A-Z]+) = (?P<id>\\d+)",
```
tmfink: You need to account for underscores, numbers, and possible after `_INS_`.
We want to have this be future-proof.

For example, you are missing `MIPS_INS_ABSQ_S`, `MIPS_INS_ADDIUR1SP`.

Reference: https://doc.rust-lang.org/stable/reference/identifiers.html

capstone-sys/build.rs
```diff
@@ -148,6 +152,62 @@ fn env_var(var: &str) -> String {
...
+
+    for cap in re.captures_iter(&bindings) {
+        impl_arch_enum.push_str(
+            &format!("impl {}_insn {{\n
+            pub fn from_insn_id(id: u32) -> {}_insn {{\n
+            match id {{\n",
+            &cap["arch"], &cap["arch"]));
+
+        // find instructions and their id
+        let re2 = Regex::new(
```
tmfink: Use descriptive names, not `re2`.

tmfink:
Thanks for the PR! I have avoided doing this kind of thing, but maybe is what we want.
I have made some review comments.

losynix:
Thank you, I updated the code following your comments except for the `TryFrom` trait implementation which requires returning a `Result` instead of a `<arch>_INS_INVALID` variant, not sure what would be more comfortable to use ?

tmfink:
> Thank you, I updated the code following your comments except for the `TryFrom` trait implementation which requires returning a `Result` (thus an `Error` instead of a `<arch>_INS_INVALID` variant), not sure what would be more comfortable to use ?

You can just use the `From` trait to convert directly, including the `<arch>_INS_INVALID`.

In the future, I think it will make sense to define a separate enum for each arch instructions without `<arch>_insns_invalid`. When conversions fail, they would be able to return `Option<ArchInsn>`.

losynix:
> You can just use the `From` trait to convert directly, including the `<arch>_INS_INVALID`.

Sure, done.

> In the future, I think it will make sense to define a separate enum for each arch instructions without `<arch>_insns_invalid`. When conversions fail, they would be able to return `Option<ArchInsn>`.

It would make sense indeed.

losynix:
everything good for you ?

capstone-sys/build.rs
```diff
@@ -40,6 +40,15 @@ extern crate bindgen;
 
 extern crate cc;
 
+#[cfg(feature = "use_bindgen")]
+extern crate regex;
```
tmfink: I should upgrade capstone-sys to the 2018 edition so that you don't need to do this

tmfink:
Thanks for the PR! Sorry for the wait.

losynix:
no problem, thank you !

PR #82, opened 2020-05-01
Allow Instructions to deref to slice of Insn

ExPixel:
Makes Insn transparent so that cs_insn can be safely converted to it.
This makes it possible to just make Instructions deref to a slice of
Insn so that is can be passed around more easily without an extra
layer of indirection on top of the slice.

This is a small quality of life improvement that allows things like this:

```rust
pub fn use_instrs<'i>(insn: &[Insn<'i>]) {
    // use as a slice
}

pub fn create_instrs() {
    let cs = Capstone::new()
        .x86()
        .mode(arch::x86::ArchMode::Mode64)
        .syntax(arch::x86::ArchSyntax::Intel)
        .detail(true)
        .build().unwrap();
    let insn = cs
        .disasm_all(symbol_code, test_symbol.addr).unwrap();

    // Deref will coerce to &[Insn<'i>] here
    use_instrs(&insn);
}

```

### DIFF
--- a/capstone-rs/src/instruction.rs
+++ b/capstone-rs/src/instruction.rs
@@ -118,7 +118,8 @@ impl<'a> core::ops::Deref for Instructions<'a> {
     type Target = [Insn<'a>];
 
     fn deref(&self) -> &[Insn<'a>] {
-        unsafe { core::slice::from_raw_parts(self.0.as_ptr() as *const Insn, self.0.len()) }
+        // SAFETY: `cs_insn` has the same memory layout as `Insn`
+        unsafe { &*(self.0 as *const [cs_insn] as *const [Insn]) }
     }
 }
 
--- a/capstone-rs/src/instruction.rs
+++ b/capstone-rs/src/instruction.rs
@@ -114,6 +114,14 @@ impl<'a> Instructions<'a> {
     }
 }
 
+impl<'a> core::ops::Deref for Instructions<'a> {
+    type Target = [Insn<'a>];
+
+    fn deref(&self) -> &[Insn<'a>] {
+        unsafe { core::slice::from_raw_parts(self.0.as_ptr() as *const Insn, self.0.len()) }
+    }
+}
+
 impl<'a> Drop for Instructions<'a> {
     fn drop(&mut self) {
         if !self.is_empty() {
@@ -184,6 +192,7 @@ impl_SliceIterator_wrapper!(
 );
 
 /// A wrapper for the raw capstone-sys instruction
+#[repr(transparent)]
 pub struct Insn<'a> {
     /// Inner `cs_insn`
     pub(crate) insn: cs_insn,
--- a/capstone-rs/src/test.rs
+++ b/capstone-rs/src/test.rs
@@ -3054,3 +3054,41 @@ fn test_arch_xcore_detail() {
         ],
     );
 }
+
+#[test]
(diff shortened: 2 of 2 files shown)


ExPixel:
Makes sense. Added a test to make sure that `cs_insn` and `Insn` always have the same size and alignment. It also checks the conversion itself.

tmfink:
@ExPixel I want you to know that I haven't forgotten about this PR--I'm just trying to convince myself that this is actually safe.

ExPixel:
Admittedly, the `slice::from_raw_parts` might be unnecessary and I might be able to get away with a couple of casts using `as` instead. I think that might actually be the recommended way of doing what amounts to the same thing and it's less error prone. I'll do that instead and update the PR.

tmfink:
> Admittedly, the `slice::from_raw_parts` might be unnecessary and I might be able to get away with a couple of (still unsafe) casts using `as` instead. I think that might actually be the recommended way of doing what amounts to the same thing and it's less error prone. I'll do that instead and update the PR.

Great--could you also rebase your changes onto master (you don' need to squash)? I fixed some of the Cirrus CI config that fixes the FreeBSD tests.

capstone-rs/src/instruction.rs
```diff
@@ -114,6 +114,14 @@ impl<'a> Instructions<'a> {
     }
 }
 
+impl<'a> core::ops::Deref for Instructions<'a> {
+    type Target = [Insn<'a>];
+
+    fn deref(&self) -> &[Insn<'a>] {
+        unsafe { core::slice::from_raw_parts(self.0.as_ptr() as *const Insn, self.0.len()) }
```
tmfink, line 121: Could you add a comment here explaining why this is safe (it doesn't have to be long).
I'm thinking something like:

```rust
// CLAIM: `Insn` has the same representation as `cs_insn` because of the `repr(transparent)` over `Insn`
```
```

**Comment by tmfink** [comment]
@ExPixel the more I think about this, the more I think we should rely on the `Deref` impl to cover what we want for `Instructions`. There's no point in defining an `.iter()` method or `InstructionIterator`.

If you want, you can try tackling the cleanup. Otherwise, I'll do it myself in a follow-up PR.

`repr(transparent)` wasn't stable when this logic was first implemented, but we may as well take advantage.  :smile:

File: capstone-rs/src/instruction.rs (hunk 1 above)
Comment (ExPixel) [reply capstone-rs/src/instruction.rs]: Didn't get a chance to see this before rebasing abd pushing my changes. I changed my implementation so it's closer to what the standard library does for similar things. Here's a similar cast in the `std::cell::Cell` implementation:

```rust
#[repr(transparent)]
pub struct Cell<T: ?Sized> {
    value: UnsafeCell<T>, // NOTE: UnsafeCell is also repr(transparent)
}

impl<T> Cell<[T]> {
    pub fn as_slice_of_cells(&self) -> &[Cell<T>] {
        // SAFETY: `Cell<T>` has the same memory layout as `T`.
        unsafe { &*(self as *const Cell<[T]> as *const [Cell<T>]) }
    }
}
```

ExPixel:
I'll look into cleaning up the uses of InstructionsIterator  as well. I'll probably make another pull request for that in the next couple of days :smile:

tmfink:
Merged as 4e07963

tmfink:
(I squashed/merged manually)

PR #85, opened 2020-05-14
Cleanup instruction iterator

ExPixel:
Removed `Instructions::iter()` and `InstructionIterator`. The nice thing about having a deref to a slice is that it gives you an `iter()` method for free because of Rust's deref coercion. Also fixed some tests so that they use references to `Insn` instead and made `Insn` cloneable as well in case someone did want an `Iterator<Item = Insn>` instead of an `Iterator<Item = &Insn>`.

### DIFF
--- a/capstone-rs/src/instruction.rs
+++ b/capstone-rs/src/instruction.rs
@@ -103,12 +103,6 @@ impl<'a> Instructions<'a> {
         self.0.len()
     }
 
-    /// Iterator over instructions
-    pub fn iter(&'a self) -> InstructionIterator<'a> {
-        let iter = self.0.iter();
-        InstructionIterator(iter)
-    }
-
     pub fn is_empty(&self) -> bool {
         self.len() == 0
     }
@@ -177,20 +171,6 @@ macro_rules! impl_SliceIterator_wrapper {
         }
     }
 }
-
-/// An iterator over the instructions returned by disasm
-///
-/// This is currently the only supported interface for reading them.
-pub struct InstructionIterator<'a>(slice::Iter<'a, cs_insn>);
-
-impl_SliceIterator_wrapper!(
-    impl<'a> Iterator for InstructionIterator<'a> {
-        type Item = Insn<'a>;
-        [
-            |x| Insn { insn: *x, _marker: PhantomData }
-        ]
-    }
-);
 
 /// A wrapper for the raw capstone-sys instruction
 #[repr(transparent)]
--- a/capstone-rs/src/instruction.rs
+++ b/capstone-rs/src/instruction.rs
@@ -38,6 +38,16 @@ pub type RegIdInt = u16;
 #[repr(transparent)]
 pub struct RegId(pub RegIdInt);
 
+impl core::convert::From<u32> for RegId {
+    fn from(v: u32) -> RegId {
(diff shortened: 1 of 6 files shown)


tmfink:
Thanks for the PR! When I have more time, I can look into this in more detail.

capstone-rs/src/instruction.rs
```diff
@@ -178,21 +172,8 @@ macro_rules! impl_SliceIterator_wrapper {
...
-    impl<'a> Iterator for InstructionIterator<'a> {
-        type Item = Insn<'a>;
-        [
-            |x| Insn { insn: *x, _marker: PhantomData }
-        ]
-    }
-);
-
 /// A wrapper for the raw capstone-sys instruction
+#[derive(Clone)]
```
tmfink, line 176: I need to think more if this is sound.
`cs_insn` has the `detail: *mut cs_detail` field.

capstone-rs/src/instruction.rs
```diff
@@ -178,21 +172,8 @@ macro_rules! impl_SliceIterator_wrapper {
     }
 }
 
-/// An iterator over the instructions returned by disasm
-///
-/// This is currently the only supported interface for reading them.
-pub struct InstructionIterator<'a>(slice::Iter<'a, cs_insn>);
-
-impl_SliceIterator_wrapper!(
```
tmfink, line 186: I think you can get rid of this macro definition as well (and all uses) if you make the same transparent/deref changes to the appropriate types.

capstone-rs/src/instruction.rs
ExPixel: This should be safe since the lifetime of `Insn` is tied to `Instructions` and that's what frees the pointer, but I might be wrong.

ExPixel:
So I changed `RegId` first and am holding off on making changes to the groups iterator. Just wanted to know what you think of the changes first.

capstone-rs/src/instruction.rs
```diff
@@ -35,8 +35,19 @@ pub type RegIdInt = u16;
 
 /// Represents an register id, which is architecture-specific.
 #[derive(Copy, Clone, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
+#[repr(transparent)]
 pub struct RegId(pub RegIdInt);
 
+impl core::convert::From<u32> for RegId {
+    fn from(v: u32) -> RegId {
+        if v <= core::u16::MAX as u32 {
```
tmfink, line 43: We should use `TryFrom` instead. Also, use `RegId` instead of `u16` directly.

```rust
RegId(x.try_into().ok().unwrap_or(0))
```

https://doc.rust-lang.org/std/primitive.u16.html#impl-TryFrom%3Cu32%3E

capstone-rs/src/instruction.rs
tmfink: I think you are right. I *think* the `Clone` should be safe. But, is it useful?

tmfink:
> So I changed `RegsIter` first and am holding off on making changes to the groups iterator. Just wanted to know what you think of the changes first.

Sorry for the ultra late comment. I think the change looks good. Please feel free to apply that same pattern to the other iterators.
If you would prefer, I could merge as-is and you can make a follow-up PR.

PR #89, opened 2020-10-20
Fix completion in CLion.

turbocool3r:
This small change doesn't affect anything but fixes completion in CLion with intellij-rust ([issue](https://github.com/intellij-rust/intellij-rust/issues/6274)).


tmfink:
Sorry for the late reply, thanks for the PR!

PR #92, opened 2021-04-04
Update vendored capstone to 71f5c64c43b9

s1341:
Update the included capstone to latest master, 71f5c64c43b9.

In particular, this fixes incompatibility with other tools due to the change in the size of the `bytes` field in `cs_insn`.


s1341:
I accidentally merged from master instead of next.... Closing this in favor of #93

PR #93, opened 2021-04-04
Update capstone to 702dbe78ca116de8e

s1341:
This is in place of #92, in which I mistakenly mistakenly merged from master, instead of next.

In particular, this removes the VESS from arm64. So I've gotten rid of that.

It also got rid of Arm64MrsReg, replacing it with Arm64Sysreg. I've made teh replacement, but there are now two failing tests. I'm unsure how to go about fixing them.

Please consider this in place of #92, which I will close.


tmfink:
Thanks for the PR, but I think I'll close this PR in favor of #88, which already fixes these issues.
I never merged it because of CI issues, but I'll try to push it through since it seems like some folks may want the C library upgraded.

tmfink:
Also, just FYI, I used the `capstone-sys/scripts/update_capstone.sh` script to update the vendored capstone.

s1341:
No worries. Does that PR get rid of VESS too?

PR #98, opened 2021-05-05
Fix link to examples/ directory in README

cxiao:
It seems that the link was broken when the capstone-rs subdirectory was added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ Found 4 instructions
     insn groups:
 ```
 
-To see more demos, see the [`examples/`](examples) directory.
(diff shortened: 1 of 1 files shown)


tmfink:
The link is *kinda* valid since `capstone-rs/README.md` is a symlink to the root `README.md`. That means the link is valid for one README but not the other. I should just delete one of them and avoid using a symlink.

cxiao:
Sounds good, if you're planning to do that I can just close this for now.

tmfink:
Thanks for pointing out the issue! I resolved with e294314c42e0572e09d987744f77e60a0780073a

cxiao:
Awesome, thanks very much!

PR #101, opened 2021-07-22
Bump bindgen version and add wasm support

s1341:
Bumped the bindgen version used (as packages with newer bindgen were conflicting).

Added WASM support, as without it I was running into issues with missing WASM related symbols.

### DIFF
--- a/capstone-sys/Cargo.toml
+++ b/capstone-sys/Cargo.toml
@@ -26,7 +26,7 @@ travis-ci = { repository = "capstone-rust/capstone-sys" }
 libc = { version = "0.2.59", default-features = false }
 
 [build-dependencies]
-bindgen = { optional = true, version = "0.53.0" }
+bindgen = { optional = true, version = "0.57.0" }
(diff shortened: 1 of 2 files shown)


tmfink:
Thanks for the PR!

s1341:
Can you spin a version so that we can benefit from this change please?

tmfink:
> Can you spin a version so that we can benefit from this change please?

Before spinning another release, I want to merge #85 and perform some related clean-up (using slices instead of iterators).

Also, I think I need check the pre-generated bindings. On my computer, the bindings are changed slightly.
@s1341 when you are building this crate, are you using the `use_bindgen` feature? Otherwise, I would not expect the `bindgen` feature bump to make a difference.

s1341:
yes. I'm using `use_bindgen`.

tmfink:
@s1341 I spun a release for `capstone-sys` 0.14.0 and `capstone` 0.10.0

s1341:
Thanks!

PR #102, opened 2021-08-02
arch: Implement size_hint for operand iterators

leoetlino:
Noticed while profiling that vector resizing was pretty hot when disassembly and getting operands, and it appears to be due to operand iterators not implementing `size_hint`. We can simply delegate size_hint to the underlying iterator.


leoetlino:
The test failures appear to be unrelated.

tmfink:
Thanks for the PR! There may be some other places where we can do a better job of passing information we already know (like this).

My plan is to expose some of the data we have a slice instead (like @ExPixel has done in #82, #85). Then, we won't need to jump through these hoops to make sure that the iterators get optimized.

Issue #1, opened 2015-12-27
Support `cs_disasm_iter`


gereeter:
Introduced in Capstone 3.0.

richo:
hey, thanks for the ping. I'll see if I can get cs_disasm_iter exposed properly as an Iterator when I get a chance.

richo:
Hey @gereeter,

I spent some time noodling on this, but I'm not sure how feasible it is without depending really directly on the implementation of slices, which is probably fine but more fiddly than I was hoping to make this.

My branch is at https://github.com/richo/capstone-rs/compare/cs_iter if you want to have a look

gereeter:
The Rust side of the API looks good to me, though it technically isn't quite as powerful as the original C api. As for avoiding implementation details, `disasm_iter` could just call `cs_malloc`, storing the allocation inside the `CsIterator`. The destructor for `CsIterator` would clean up that allocation, and that allocation would be passed to every call to `cs_disasm_iter`.

gereeter:
Actually, I just remembered that copying the instruction is probably not valid, because of the detail pointer - copying the instruction would not copy its detail, so any further iterations would then overwrite the detail, since all instructions produced would be pointing at the same `cs_detail` struct.

I think this means that the `CsIterator` needs to be some form of streaming iterator that only allows access to the current instruction and doesn't allow looking at multiple instructions at once - this also more closely matches the use case described in the Capstone docs. For example:

```rust
pub struct InsnIter<'a> {
    handle: &'a Capstone,
    ptr: Unique<Insn>
}

impl<'a> Drop for InsnIter<'a> {
    fn drop(&mut self) {
        cs_free(*self.ptr, 1);
    }
}

impl InsnIter<'a> {
    pub fn new(handle: &'a Capstone) -> InsnIter<'a> {
        InsnIter {
            handle: handle,
            ptr: Unique::new(cs_malloc(handle.csh))
        }
    }

    pub fn next(&mut self, code: &[u8], address: u64) -> CsResult<&Insn>;
}
```

richo:
I don't think I need to wrap `cs_malloc` at all- it's called under the hood by `cs_disasm_iter` according to the docs. What's more plausible is that I'd want to shove jemalloc in iff it's a rust library that contains the entry point.

gereeter:
[...]

PR #114, opened 2021-09-19
adding disasm_iter functionality

dnsserver:
I needed the recursive disasm functionality. 
I haven't done extensive testing although I've been using this code for some time now.

### DIFF
--- a/capstone-rs/examples/recursive.rs
+++ b/capstone-rs/examples/recursive.rs
@@ -62,7 +62,7 @@ fn main() {
     let mut addr_queue: VecDeque<u64> = VecDeque::new();
     let mut addr_seen: HashMap<u64, bool> = HashMap::new();
 
-    let cs_ins = cs.iter_malloc();
+    let cs_ins = cs.malloc();
     assert!(!cs_ins.is_null());
 
     addr_queue.push_back(obj.entry());
@@ -96,4 +96,6 @@ fn main() {
             }
         }
     }
+
+    cs.free(cs_ins);
 }
--- a/capstone-rs/examples/recursive.rs
+++ b/capstone-rs/examples/recursive.rs
@@ -68,26 +68,22 @@ fn main() {
     addr_queue.push_back(obj.entry());
 
     while !addr_queue.is_empty() {
-        let mut qaddr = addr_queue.pop_front().unwrap();
-        if let Some(_) = addr_seen.get(&qaddr) {
+        let mut addr = addr_queue.pop_front().unwrap();
+        if let Some(_) = addr_seen.get(&addr) {
             continue;
         }
-        addr_seen.insert(qaddr, true);
+        addr_seen.insert(addr, true);
 
-        println!("addr: {:#02x?}", qaddr);
+        println!("addr: {:#02x?}", addr);
 
-        let offset = (qaddr - sec_addr) as usize;
-        let qpc = &mut sec_text[offset..].as_ptr();
-        let mut qcount = sec_text.len() - offset;
-
-        println!("output: count:{} addr:{:#02x?}", qcount, qaddr);
+        let mut offset = (addr - sec_addr) as usize;
         loop {
-            let (ret, count, addr) = cs.disasm_iter(qpc, qcount, qaddr, cs_ins);
+            let (ret, o, a) = cs.disasm_iter(&sec_text, offset, addr, cs_ins);
             if !ret {
                 break;
             }
-            qcount = count;
-            qaddr = addr;
+            offset = o;
+            addr = a;
 
             let ins = unsafe { Insn::from_raw(cs_ins) };
             println!("{}", ins);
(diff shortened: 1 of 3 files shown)


capstone-rs/examples/recursive.rs
```diff
@@ -0,0 +1,98 @@
...
+		eprintln!("no section .text found");
+		process::exit(2);
+	};
+
+    println!(".text section addr: {:#02x?} size: {:#02x?}", sec_addr, sec_text.len());
+    
+    let mut addr_queue: VecDeque<u64> = VecDeque::new();
+    let mut addr_seen: HashMap<u64, bool> = HashMap::new();
+
+    let cs_ins = cs.iter_malloc();
```
tmfink, line 60: I *think* this is undefined behavior (instantiating an Rust struct with initialized memory).

To work with uninitialized data, look into the [MaybeUninit](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html) type.

tmfink (changes requested):

tmfink:
Thanks for the PR! What is the purpose of `disasm_iter()` here? What issues were you running into?

capstone-rs/src/capstone.rs
```diff
@@ -165,6 +165,28 @@ impl Capstone {
         }
     }
 
+    pub fn iter_malloc<'a>(&'a self) -> *mut cs_insn {
+        unsafe {
+            cs_malloc(self.csh())
```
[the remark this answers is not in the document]
dnsserver: 100%! I will update, thanks.

dnsserver:
I didn't run into any issues. I needed to do recursive disassembly for obfuscated instructions.

capstone-rs/examples/recursive.rs
dnsserver: I will double check this. However the iter_malloc does only initialize the capstone handle, which is a raw pointer, not a struct:
`csh: *mut c_void` .

tmfink:
> I didn't run into any issues. I needed to do recursive disassembly for obfuscated instructions.

Sorry I wasn't clear with my question. What did this crate not allow you to do before that this new function `disasm_iter()` allows you to do?

dnsserver:
Oh, recursive disassembly is good for use cases where data and code are mixed or when x86 instructions are obfuscated.
This is what this issue is about: https://github.com/capstone-rust/capstone-rs/issues/1
Pretty much allows me to disassemble chunks of bytes at a time, specifying offset where to start disassembling (following jumps), and have control when to stop. This is much easier done with capstone then trying to manually feed the chunks of data to disasm_all function IMHO.

tmfink (changes requested):
I'm happy to add an example program that does recursive disassembly, but I'm not inclined to accept the new `disasm_iter()` function in the current state since it allows users to violate memory safety.

One of the biggest advantages of Rust is the memory safety, so we want to make sure we preserve that.

You can see from #1 that I've previously tried to add support for `cs_disasm_iter()` unsuccessfully.

My recommendation is that you remove the `disasm_iter()` feature from this PR and just add the recursive disassembly example. If you want to in a follow-up PR, you can try to add `disasm_iter()` in a memory-safe way.

capstone-rs/src/capstone.rs
```diff
@@ -165,6 +165,37 @@ impl Capstone {
...
+
+    pub fn free<'a>(&'a self, insn: *mut cs_insn) {
+        unsafe {
+            cs_free(insn, 1);
+        }
+    }
+
+    pub fn disasm_iter<'a>(
+        &'a self,
+        code: *mut *const u8,
```
tmfink, line 180: Avoid exposing raw pointers in the public API. We want to expose a memory-safe interface.

dnsserver: Thank you for all your feedback. I really appreciate it.
I will take a look and see if I can clean even more the API. Feel free to decline this PR.

capstone-rs/src/capstone.rs
```diff
@@ -165,6 +165,37 @@ impl Capstone {
...
+        unsafe { cs_malloc(self.csh()) }
+    }
+
+    pub fn free<'a>(&'a self, insn: *mut cs_insn) {
+        unsafe {
+            cs_free(insn, 1);
+        }
+    }
+
+    pub fn disasm_iter<'a>(
```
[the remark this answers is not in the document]
tmfink: I need to think more about what the `disasm_iter()` function interface should be.

Instead of a C-like interface, we want a "Rusty" interface that:

1. Does not require the user to use unsafe
    a. Does not take any pointers
2. Does not require manual malloc/free functions to avoid memory leaks, double frees, etc.
3. Is easy to use
    a. Does not return a tuple (an enum/struct instead)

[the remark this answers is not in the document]
dnsserver: I agree. Let me close this PR. I can do a followup with these guidelines

PR #115, opened 2021-09-21
Support cs_disasm_iter

dnsserver:
What do you think about this API?


tmfink:
This is a big improvement over the last PR.  :smile:

Overall, I like the approach of having separate type to "own" the `cs_insn` allocation. I also like how `DisasmIter` tracks the "updates" from repeated calls to `cs_disasm_iter()`.

I have some comments that I will make in-line.

capstone-rs/examples/recursive.rs
```diff
@@ -0,0 +1,87 @@
...
+    let args: Vec<_> = env::args().collect();
+    if args.len() != 2 {
+        eprintln!("Usage: {} <file>", args[0]);
+        process::exit(-1);
+    }
+
+    let buf = if let Ok(bd) = fs::read(&args[1]) {
+        bd
+    } else {
+        eprintln!("cannot read file");
```
tmfink, line 22: Since this is an example program and not a library, you can just use `.expect("my message")` on `Option`/`Result` instead. This will keep the example shorter and easier to read.

capstone-rs/examples/recursive.rs
```diff
@@ -0,0 +1,87 @@
...
+
+    let obj = if let Ok(od) = object::File::parse(&*buf) {
+        od
+    } else {
+        eprintln!("cannot parse file");
+        process::exit(-1);
+    };
+
+    let mut addr_queue: VecDeque<u64> = VecDeque::new();
+    let mut addr_seen: HashMap<u64, bool> = HashMap::new();
```
tmfink, line 34: You can use a `HashSet` instead

capstone-rs/examples/recursive.rs
```diff
@@ -0,0 +1,87 @@
...
+    for section in obj.sections() {
+        if section.kind() == SectionKind::Text {
+            println!("{:x?} ", section);
+            addr_queue.push_back(section.address());
+        }
+    }
+
+    let cs = if let Ok(cs) = Capstone::new()
+        .x86()
+        .mode(arch::x86::ArchMode::Mode64)
```
tmfink, line 45: You should add a module comment `//!` and update the help message to specify that this program expects amd64 AKA "x86_64" binaries.

capstone-rs/examples/recursive.rs
```diff
@@ -0,0 +1,87 @@
...
+    {
+        cs
+    } else {
+        eprintln!("failed to create capstone handle");
+        process::exit(-1);
+    };
+
+    let mut disasm = cs.get_disasm_iter();
+
+    while !addr_queue.is_empty() {
```
tmfink, line 57: Instead of using `pop_front().unwrap()` inside the loop, you can use "while let" like:

```
while let Some(addr) = addr_queue.pop_front() {
    // ...
}
```

This is more efficient and avoids the need to `unwrap()`.

capstone-rs/examples/recursive.rs
```diff
@@ -0,0 +1,87 @@
...
+        println!("---> addr: {:#02x?}", addr);
+
+        let offset = addr as usize;
+        let mut cur_insn = disasm.disasm_iter(&buf, offset, addr);
+        while let Ok(insn) = cur_insn {
+            if insn.id() == InsnId(X86Insn::X86_INS_INVALID as u32) {
+                break;
+            }
+            println!("{}", insn);
+            match X86Insn::from(insn.id().0) {
```
tmfink, line 73: Pull this match statement into a help function like `insn_ends_basic_block()`.

Also, instead of looking for specific instruction IDs, you can match on the instruction groups
https://docs.rs/capstone/0.10.0/capstone/struct.InsnDetail.html#method.groups

capstone-rs/src/capstone.rs
```diff
@@ -165,6 +165,18 @@ impl Capstone {
         }
     }
 
+    /// Creates an instance of DisasmIter structure
+    ///
+    pub fn get_disasm_iter<'a>(&'a self) -> DisasmIter<'a> {
+        DisasmIter {
+            insn: unsafe { cs_malloc(self.csh()) },
```
tmfink, line 172: we should panic on allocation failures, per usual Rust norms (currently this will fail silently)

capstone-rs/src/capstone.rs
```diff
@@ -423,3 +435,64 @@ impl Drop for Capstone {
...
+    ///
+    /// usage shown in examples/recursive.rs
+    pub fn disasm_iter_continue(&mut self, code: &[u8]) -> CsResult<Insn> {
+        self.disasm_iter(code, self.offset, self.addr)
+    }
+
+    /// Used to start the iterative disassembly
+    ///
+    /// usage shown in examples/recursive.rs
+    pub fn disasm_iter(&mut self, code: &[u8], offset: usize, addr: u64) -> CsResult<Insn> {
```
tmfink, line 468: we don't need to take an offset; the user can just adjust the `code` slice as they want:

```rust
let new_code = &code[3..];
```

capstone-rs/src/capstone.rs
```diff
@@ -423,3 +435,64 @@ impl Drop for Capstone {
...
+        let code_len = code.len();
+        let code_ptr = &mut code[offset..].as_ptr();
+        let mut c = code_len - offset;
```
tmfink, line 471: Avoid one-word variable names. Perhaps something like `new_code_ptr` and `new_addr`?

tmfink (changes requested):

capstone-rs/src/capstone.rs
dnsserver: It will get a bit tricky to handle the iteration within the loop without an offset ( `cur_insn = disasm.disasm_iter_continue(&buf);` ). Would it be too much to ask to keep the offset ?

capstone-rs/src/capstone.rs
dnsserver: I couldn't find an example I could follow in this repo for the handling a malloc issue. Can you please point me to an example if what I pushed now is not the right approach? Thnx!

tmfink: Nothing too crazy--if the return for `cs_malloc()` is NULL, then just call `panic!("cs_malloc() failed")`. :-)

tmfink:
I just pushed a fix (#116) for the GitHub action failure; please rebase to the current `master` branch to get the fix.

jiegec:
I would prefer to create a rust-style Iterator that wraps the `cs_disasm_iter` logic so that the user can iterate instructions quickly, and provide a wrapper for the raw `cs_disasm_iter` function if the user wants to customize the usage.

capstone-rs/src/capstone.rs
```diff
@@ -205,6 +205,43 @@ impl Capstone {
...
+    /// # let cs = Capstone::new().x86().mode(arch::x86::ArchMode::Mode32).build().unwrap();
+    /// let mut iter = cs.disasm_iter(b"\x90", 0x1000).unwrap();
+    /// assert_eq!(iter.next().unwrap().mnemonic(), Some("nop"));
+    /// assert!(iter.next().is_none());
+    /// ```
+    ///
+    /// # Errors
+    ///
+    /// If `cs_malloc` failed due to OOM, [`Err(Error::OutOfMemory)`](Error::OutOfMemory) is returned.
+    pub fn disasm_iter<'a, 'b>(
```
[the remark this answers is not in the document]
jiegec: Nice idea!

PR #117, opened 2021-09-29
Fix typo 'lift' to 'left'

rfalke:


tmfink:
Thanks for the PR! I was probably sleep-deprived when I first added ARM support. :smile:

PR #120, opened 2022-03-21
Add cs_regs_access support

JimColerick:
Added support for the cs_regs_access function and implemented a corresponding struct to contain the data returned by the function.

### DIFF
--- a/capstone-rs/src/instruction.rs
+++ b/capstone-rs/src/instruction.rs
@@ -12,7 +12,7 @@ use crate::arch::ArchDetail;
 use crate::constants::Arch;
 use crate::ffi::str_from_cstr_ptr;
 
-pub type cs_regs = [u16; 64];
+pub type CsRegs = [u16; 64];
 
 /// Represents a slice of [`Insn`] returned by [`Capstone`](crate::Capstone) `disasm*()` methods.
 ///
@@ -388,7 +388,8 @@ impl<'a> Display for Instructions<'a> {
         Ok(())
     }
 }
-
+/// Registers read and/or written to by a given instruction
+#[derive(Debug)]
 pub struct RegAccess {
     /// `Vec` containing the `RegId` of the registers read from
     read: Vec<RegId>,
--- a/capstone-rs/src/instruction.rs
+++ b/capstone-rs/src/instruction.rs
@@ -4,12 +4,15 @@ use core::marker::PhantomData;
 use core::ops::Deref;
 use core::slice;
 use core::str;
+use alloc::vec::Vec;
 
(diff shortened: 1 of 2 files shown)


richo:
This looks good to me, but in truth I've been away from the project for a while 😅

I'm planning to wait for one of the more active folks to chime in

capstone-rs/src/instruction.rs
```diff
@@ -4,13 +4,16 @@ use core::marker::PhantomData;
...
 use core::str;
+use alloc::vec::Vec;
 
 use capstone_sys::*;
 
 use crate::arch::ArchDetail;
 use crate::constants::Arch;
 use crate::ffi::str_from_cstr_ptr;
 
+pub type CsRegs = [u16; 64];
```
tmfink, line 15: Remove this since there's no need to redefine the `cs_regs` type. This type does not need to be public.

capstone-rs/src/instruction.rs
```diff
@@ -385,6 +388,44 @@ impl<'a> Display for Instructions<'a> {
...
+    pub(crate) fn from_slice_u16(
+        registers_read: &CsRegs,
+        registers_read_count: usize,
+        registers_written: &CsRegs,
+        registers_written_count: usize,
+        ) -> Self {
+            let read = registers_read[0..registers_read_count]
+                .iter()
+                .cloned()
+                .map(|x| RegId::from(x as u32))
```
tmfink, line 411: since `RegIdInt` is defined to be `u16`, could you define a new impl `From<u16>` for `RegId`? There's no need to cast here.

You could already just do `RegId(x)` (since the field is public).

capstone-rs/src/instruction.rs
```diff
@@ -385,6 +388,44 @@ impl<'a> Display for Instructions<'a> {
         Ok(())
     }
 }
+/// Registers read and/or written to by a given instruction
+#[derive(Debug)]
+pub struct RegAccess {
+    /// `Vec` containing the `RegId` of the registers read from
+    read: Vec<RegId>,
```
tmfink, line 395: we can avoid the 2 heap allocations by storing the `cs_regs` in this struct along with their lengths:

```rust
pub struct RegAccess {
    regs_read: cs_regs,
    regs_read_count: u8,
    regs_write: cs_regs,
    regs_write_count: u8,
}
```

You can add methods to access the data safely as a slice reference.

See `InsnDetail::regs_read()` in `instruction.rs` for an example.

tmfink (changes requested):
Thanks for the PR! Sorry for running late on the review.

PR #123, opened 2022-04-01
Introduce an owned insn type

### LINKED ISSUE #121: UAF when `clone`ing an `Insn` with a `detail` pointer
While iterating over `operands` (from the `X86InsnDetail` binding) I'm experiencing a crash due to an invalid `op_type`.

The crash occurs specifically in https://github.com/capstone-rust/capstone-rs/blob/fa37a3779b440fe9efe4f30265a714b09ea77076/capstone-rs/src/arch/x86.rs#L30-L42 - `op_type` has a value of `1633878019`  which causes an invalid read from the `match`'s jump table.

If I force my program to skip the offending instruction, it later panics on an out-of-bounds array access: https://github.com/capstone-rust/capstone-rs/blob/fa37a3779b440fe9efe4f30265a714b09ea77076/capstone-rs/src/arch/mod.rs#L671, where the `operands` array is of size 8 and `op_count` has a value of 160.

Neither crash occurs when running under Valgrind, so I haven't been able to trace the root cause of the memory corruption.

I'm not sure if this is a bug in capstone-rs, capstone-sys, or Capstone itself, but I'm reporting it here as that's where the crash is occuring for me.

Unfortunately, I haven't been able to create a reliable reproducer, as the crash is only triggered in a large program. However, this snippet demonstrates the configuation I am using and code I am disassembling. The crash occurs at the instruction at offset 0x30: `0x2030: jmp qword ptr [rip + 0x6e4a]`
```rs
static CODE: &[u8; 112] = b"\xf3\x0f\x1e\xfaH\x83\xec\x08H\x8b\x05Ao\x00\x00H\x85\xc0t\x02\xff\xd0H\x83\xc4\x08\xc3\x00\x00\x00\x00\x00\xff5Jn\x00\x00\xff%Ln\x00\x00\x0f\x1f@\x00\xff%Jn\x00\x00h\x00\x00\x00\x00\xe9\xe0\xff\xff\xff\xff\x15Jn\x00\x00\xff\x15Dn\x00\x00P\xff\x15=n\x00\x00\xff\x157n\x00\x00\xff\x151n\x00\x00\xff\x15+n\x00\x00\xff\x15%n\x00\x00\xff\x15\x1fn\x00";
use capstone::prelude::*;
use capstone::arch;
use capstone::arch::x86::{
    X86Insn, X86InsnDetail, X86OpMem, X86OperandType,
    X86Reg::{X86_REG_EIP, X86_REG_RIP},
};
use capstone::arch::ArchDetail;
use capstone::{prelude::*, Capstone, Insn, RegAccessType};
fn main() {
    let cs = Capstone::new()
        .x86()
        .mode(arch::x86::ArchMode::Mode64)
        .syntax(arch::x86::ArchSyntax::Intel)
        .detail(true)
        .build()
        .expect("Couldn't initialise Capstone");
    let instructions = cs.disasm_all(CODE, 0x2000).unwrap();
    let instructions: Vec<_> = instructions.iter().collect();
    for i in instructions {
        println!("{}", i);
        if i.mnemonic().unwrap().starts_with('j') {
            let detail = cs.insn_detail(i).unwrap();
            if let ArchDetail::X86Detail(detail) = detail.arch_detail() {
                let res = detail.operands().find_map(|op| {
                    if let X86OperandType::Mem(mem) = op.op_type {
                        // We only accept rip-relative addresses
                        if (mem.base().0 as u32 != X86_REG_EIP && mem.base().0 as u32 != X86_REG_RIP)
                            || mem.index() != RegId::INVALID_REG
                        {
                            None
                        } else {
                            Some((mem, op.access.unwrap()))
                        }
                    } else {
                        None
                    }
                });
                dbg!(res);
            }
        }
    }
    
}
```

clubby789:
As mentioned in #121 and #122, cloning an `Insn` followed by dropping the `Instructions` slice it belongs to causes a use-after-free. This PR makes two changes

- Adding `no_copy` to bindgen
This prevents the `cs_insn` structure as being marked safe to `Copy`. I also removed the `Clone` implementation as it used the buggy `Copy`

- Introducing `OwnedInsn`
I refactored the functionality of `Insn` into an `Instruction` trait (added to the prelude) and added a method to convert an `&Insn` into an `OwnedInsn`. This copies the data onto the Rust heap, and has a `drop` implementation that frees it from the Rust heap

Closes #121


capstone-rs/src/instruction.rs
```diff
@@ -157,12 +159,54 @@ impl<'a> Drop for Instructions<'a> {
     }
 }
 
+pub trait Instruction {
```
danielhenrymantilla, line 162:
```suggestion
pub unsafe trait Instruction {
```

or seal it; since an evil implementation could provide a silly `.insn()` implementation

capstone-rs/src/instruction.rs
```diff
@@ -172,6 +216,51 @@ pub struct Insn<'a> {
...
+    }
+}
+
+impl<'a> From<&Insn<'_>> for OwnedInsn<'a> {
+    fn from(insn: &Insn<'_>) -> Self {
+        let mut new = MaybeUninit::<cs_insn>::uninit();
+        let mut new: cs_insn = unsafe {
+            core::ptr::copy(&insn.insn as *const cs_insn, new.as_mut_ptr(), 1);
+            new.assume_init()
+        };
```
danielhenrymantilla, line 231:
```suggestion
        let mut new = <*const cs_insn>::copy(&insn.insn);
```

capstone-rs/src/instruction.rs
```diff
@@ -172,6 +216,51 @@ pub struct Insn<'a> {
...
+        new.detail =  if new.detail as usize == 0 {
+            new.detail
+        } else {
+            unsafe {
+                let new_detail = alloc::alloc::alloc(Layout::new::<cs_detail>()) as *mut cs_detail;
+                core::ptr::copy(new.detail, new_detail, 1);
+                new_detail
+            }
+        };
```
danielhenrymantilla, line 240:
```suggestion
        new.detail =  if new.detail.is_null() {
            new.detail
        } else {
            let new_detail = Box::new(*new.detail);
            new.detail = Box::into_raw(new_detail);
        };
```

capstone-rs/src/instruction.rs
```diff
@@ -203,66 +292,45 @@ pub struct Insn<'a> {
...
-    ///
-    /// Be careful this is still in early stages and largely untested with various `cs_option` and
-    /// architecture matrices
-    pub(crate) unsafe fn detail(&self, arch: Arch) -> InsnDetail {
-        InsnDetail(&*self.insn.detail, arch)
+impl<'a> Drop for OwnedInsn<'a> {
+    fn drop(&mut self) {
+        unsafe {
+            alloc::alloc::dealloc(self.insn.detail as *mut u8, Layout::new::<cs_detail>());
+        }
```
danielhenrymantilla, line 329:
```suggestion
        if let Some(ptr) = ptr::NonNull::new(self.insn.detail) {
            unsafe {
                drop(Box::from_raw(ptr.as_ptr()));
            }
        }
```

capstone-rs/src/capstone.rs
```diff
@@ -381,7 +381,7 @@ impl Capstone {
     /// 1. Instruction was created with detail enabled
     /// 2. Skipdata is disabled
     /// 3. Capstone was not compiled in diet mode
-    pub fn insn_detail<'s, 'i: 's>(&'s self, insn: &'i Insn) -> CsResult<InsnDetail<'i>> {
+    pub fn insn_detail<'s, 'i: 's>(&'s self, insn: &'i impl Instruction) -> CsResult<InsnDetail<'i>> {
```
danielhenrymantilla, line 384: It may be possible to still be using `Insn<'_>` here, and just have `OwnedInsn<'captstone>` deref to `Insn<'capstone>` through a pointer cast. That ought to alleviate the resulting ergonomics / API readability quite a bit 🙂

danielhenrymantilla (changes requested):
Having a separate type using Rust ownership / and Rust allocation (at least until capstone decides to expose its own `cs_mem_alloc` and `cs_mem_free` functions) seems indeed like the best way to replace the to-be-removed `Clone` functionality ✅ 

The suggested implementation in this PR needs to change, however, to use `drop(Box` on the `detail` over `cs_free`, since the newly resulting `cs_insn` of an `OwnedInstr` is not only not-necessarily heap-allocated, but its `detail` pointer is `Box` / `alloc::alloc`-allocated, which may use a different allocator than the non-exposed `cs_mem_malloc`.

danielhenrymantilla (approved):
👌

clubby789:
I think the tests are failing since I removed the Insn::from_raw. I don't think there's a safe way to create an Insn out of a *cs_insn that wouldn't end up replacing being incorrectly freed.

Should this be replaced with OwnedInsn::from_raw?

capstone-rs/src/instruction.rs
```diff
@@ -172,6 +225,56 @@ pub struct Insn<'a> {
...
+    }
+}
+
+impl<'a> From<&Insn<'_>> for OwnedInsn<'a> {
+    fn from(insn: &Insn<'_>) -> Self {
+        let mut new = unsafe { <*const cs_insn>::read(&insn.insn as _) };
+        new.detail = if new.detail.is_null() {
+            new.detail
+        } else {
+            unsafe {
```
tmfink, line 240: Add a comment like this:

```rust
// SAFETY: assumes that `cs_detail` struct transitively only contains owned
// types and no pointers, including the union over the architecture-specific
// types.
```

I looked over the `cs_detail` definition and it seems fine, but we need to make sure that upstream capstone doesn't make "interesting" changes with how `cs_detail` is stored. I've fixed several soundness issues upstream where something weird was being done to shared mutable state without synchronization.

Changing `cs_detail` would be an API/ABI break, so we *should* be fine with upgrades to new minor versions of capstone.

tmfink:
Thanks for the carefully constructed PR! It will take a little time to review this carefully.

capstone-rs/src/instruction.rs
```diff
@@ -157,12 +158,64 @@ impl<'a> Drop for Instructions<'a> {
...
 }
 
+/// Shared functionality for structs wrapping [`cs_insn`]
+///
+/// # Safety
+///
+/// The [`cs_insn::detail`] pointer must be safely allocated and correspondingly freed
+/// to implement this trait safely.
+#[allow(clippy::len_without_is_empty)]
+pub unsafe trait Instruction {
```
tmfink, line 168: Since you already `impl<'a> Deref for OwnedInsn<'a>`, is the trait necessary? It exposes methods that we don't want public like `insn()` and `detail()`.

There is no public trait in `std` for all `str` methods; `String` just derefs to `str`.

capstone-rs/src/instruction.rs
```diff
@@ -172,6 +225,56 @@ pub struct Insn<'a> {
     pub(crate) _marker: PhantomData<&'a InsnDetail<'a>>,
 }
```
tmfink, line 226: could you add `from_raw()` back as a method for `Insn`?

capstone-rs/src/instruction.rs
```diff
@@ -203,66 +306,45 @@ pub struct Insn<'a> {
...
+        if let Some(ptr) = core::ptr::NonNull::new(self.insn.detail) {
+            unsafe { drop(Box::from_raw(ptr.as_ptr())) }
+        }
     }
 }
 
-impl<'a> Debug for Insn<'a> {
+impl<'a> Debug for OwnedInsn<'a> {
     fn fmt(&self, fmt: &mut Formatter) -> Result<(), Error> {
         fmt.debug_struct("Insn")
```
tmfink, line 349: just `deref()` to `Insn` and call its method

capstone-rs/src/instruction.rs
```diff
@@ -273,8 +355,7 @@ impl<'a> Debug for Insn<'a> {
             .finish()
     }
 }
-
-impl<'a> Display for Insn<'a> {
+impl<'a> Display for OwnedInsn<'a> {
     fn fmt(&self, fmt: &mut Formatter) -> fmt::Result {
```
tmfink, line 359: just `deref()` to `Insn` and call its method

capstone-rs/src/test.rs
```diff
@@ -3198,20 +3198,3 @@ fn test_insn_size_and_alignment() {
     }
 }
 
-#[test]
```
tmfink, line 3201: add these back once from_raw() is added back

capstone-rs/src/instruction.rs
```diff
@@ -203,66 +306,45 @@ pub struct Insn<'a> {
...
-
-    /// Access instruction id
-    pub fn id(&self) -> InsnId {
-        InsnId(self.insn.id)
-    }
-
-    /// Size of instruction (in bytes)
-    fn len(&self) -> usize {
-        self.insn.size as usize
+impl<'a> Drop for Insn<'a> {
```
tmfink, line 309: This impl should not be needed since `impl<'a> Drop for Instructions<'a>` handles this.

tmfink (changes requested):

tmfink:
> I think the tests are failing since I removed the Insn::from_raw. I don't think there's a safe way to create an Insn out of a *cs_insn that wouldn't end up replacing being incorrectly freed.
> 
> Should this be replaced with OwnedInsn::from_raw?
> 
> Edit: One way to make it safe would be to copy only the plain data portion of `cs_insn`, and leave the `detail` pointer as null.

There should be no `Drop` impl for `Isns`, so there should be no issue. The `Instructions` `Drop` impl frees the instructions normally. If you provide an "orphan" `cs_insn` to `Insn::from_raw()`, then you are responsible for freeing the backing allocation.

It would be good to add that to the existing `from_raw()` doc comment:
```rust
/// ...
/// The caller is fully responsible for the backing allocations lifetime, including freeing.
```

capstone-rs/src/instruction.rs
danielhenrymantilla: Good catch; it could actually be interesting to use this drop impl to write a `cfg(debug_assertions)` impl of `Drop` that calls `unreachable!()`

tmfink: Interesting idea--but I don't think we want to implement `Drop` because it would affect some borrowing ergonomics because of the [drop check](https://doc.rust-lang.org/nomicon/dropck.html).

clubby789:
I think I've made all the changes you asked for, as well as adding a new test. Split into individual commits for easier review, feel free to squash them if you merge 🙂

capstone-rs/src/instruction.rs
danielhenrymantilla: Precisely because it represents unreachable code, drop check should affect no actual code whatsoever. Still, I reckon that a `cfg`-gated `unreachable!` inside a drop impl seems to push the debug assertions thing a bit too far 😅

capstone-rs/src/instruction.rs
```diff
@@ -214,53 +171,153 @@ impl<'a> Insn<'a> {
...
+    ///
+    /// # Safety
+    /// The [`cs_insn::detail`] pointer must be valid and non-null.
     pub(crate) unsafe fn detail(&self, arch: Arch) -> InsnDetail {
-        InsnDetail(&*self.insn.detail, arch)
+        InsnDetail(&*self.insn().detail, arch)
+    }
+
+    // Any pointers are owned by the Capstone C library and should be freed accordingly
+    #[inline(always)]
```
tmfink, line 226: Is there a benefit to add an getter method here? I think it may be simpler (and cause fewer lines to change) if you get rid of this method.

capstone-rs/src/instruction.rs
```diff
@@ -214,53 +171,153 @@ impl<'a> Insn<'a> {
...
+/// To learn how to get more instruction details, see [`InsnDetail`].
+pub struct OwnedInsn<'a> {
+    /// Inner cs_insn
+    pub(crate) insn: cs_insn,
+
+    /// Adds lifetime
+    pub(crate) _marker: PhantomData<&'a InsnDetail<'a>>,
+}
+
+/// Contains architecture-independent details about an [`Insn`].
```
tmfink, line 290: could you hoist this back up higher in the file (where it was before) to simplify this change?

capstone-rs/src/instruction.rs
```diff
@@ -214,53 +171,153 @@ impl<'a> Insn<'a> {
...
+
+    // Any pointers are owned by the Capstone C library and should be freed accordingly
+    #[inline(always)]
+    fn insn(&self) -> &cs_insn {
+        &self.insn
     }
 }
 
+/// A single disassembled CPU instruction.
+///
```
tmfink, line 233: could you hoist this back to its original location?

capstone-rs/src/instruction.rs
clubby789: This was a leftover from implementing `Instruction` as a trait, good catch

clubby789:
I'm not planning to implement it in this PR, but in the future it might help to have an `OwnedInstructions` type (`pub struct Instructions<'a>(&'a mut [OwnedInstruction]);`). An `Instructions` could be converted to `OwnedInstructions` by copying the whole block of memory into one large box, then copying each detail pointer into its own box.

With the current implementation, each `Insn` is boxed individually, which means far more allocations for a similar amount of memory.

capstone-rs/src/instruction.rs
```diff
@@ -256,11 +260,58 @@ impl<'a> Insn<'a> {
...
+/// 1. [`OwnedInsn`] and [`Insn`] must be `#repr(transparent)` of [`cs_insn`]
+/// 2. all [`Insn`] methods must be safe to perform for an [`OwnedInsn`]
+impl<'a> Deref for OwnedInsn<'a> {
+    type Target = Insn<'a>;
+    fn deref(&self) -> &Self::Target {
+        unsafe { &*(&self.insn as *const cs_insn as *const Insn) }
+    }
+}
+
+/// A single disassembled CPI instruction that lives on the Rust heap.
```
tmfink, line 302: typo, should be "CPU"

tmfink (changes requested):

PR #126, opened 2022-04-22
Some performance changes

clubby789:
This PR mostly just inlines property accesses, which gave performance improvements of up to 4-6x in my own project.

I also switched to using `from_utf8_unchecked` as the function requires valid UTF8 to be passed anyway. This gave some small performance improvements.

### DIFF
--- a/capstone-rs/src/ffi.rs
+++ b/capstone-rs/src/ffi.rs
@@ -11,13 +11,14 @@ use libc::{self, c_char};
 ///   lifetime appropriately
 #[inline]
 pub(crate) unsafe fn str_from_cstr_ptr<'a>(ptr: *const c_char) -> Option<&'a str> {
-    (!ptr.is_null()).then(|| {
-        let len = libc::strlen(ptr);
-
-        /* ASSUMPTION: capstone returns NUL terminated string */
-        let view: &[u8] = slice::from_raw_parts(ptr as *const u8, len as usize);
-        str::from_utf8_unchecked(view)
-    })
+    if ptr.is_null() {
+        return None;
+    }
+    let len = libc::strlen(ptr);
+    /* ASSUMPTION: capstone returns NUL terminated string */
+    let view: &[u8] = slice::from_raw_parts(ptr as *const u8, len as usize);
+    /* ASSUMPTION: capstone returns a valid UTF-8 string */
+    Some(str::from_utf8_unchecked(view))
 }
 
 #[cfg(test)]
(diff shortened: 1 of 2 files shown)


capstone-rs/src/ffi.rs
```diff
@@ -6,19 +6,18 @@ use libc::{self, c_char};
...
 /// - This function "creates" a reference with an arbitrary lifetime, so be careful to limit the
 ///   lifetime appropriately
+#[inline]
 pub(crate) unsafe fn str_from_cstr_ptr<'a>(ptr: *const c_char) -> Option<&'a str> {
-    if ptr.is_null() {
-        return None;
-    }
-
-    let len = libc::strlen(ptr);
+    (!ptr.is_null()).then(|| {
```
tmfink, line 14: I think the original if statement is more clear than `bool::then`

capstone-rs/src/ffi.rs
```diff
@@ -6,19 +6,18 @@ use libc::{self, c_char};
...
+        let len = libc::strlen(ptr);
 
-    /* ASSUMPTION: capstone returns NUL terminated string */
-    let view: &[u8] = slice::from_raw_parts(ptr as *const u8, len as usize);
-    str::from_utf8(view).ok()
+        /* ASSUMPTION: capstone returns NUL terminated string */
+        let view: &[u8] = slice::from_raw_parts(ptr as *const u8, len as usize);
+        str::from_utf8_unchecked(view)
```
tmfink, line 19: add a comment above this line explaining that it is assumed that `view` is valid UTF-8

tmfink, line 19: add a comment above this line explaining that it is assumed that `view` is valid UTF-8

capstone-rs/src/ffi.rs
tmfink: Also, try to minimize the number changed lines to reduce the size of the change (where it makes sense)

tmfink (changes requested):

tmfink:
Great work on the change! Looks like we got about 2% improvement on our existing benchmarks (although we should add more benchmarks).

```
$ cargo bench --bench my_benchmark
disasm_x86              time:   [7.5322 ms 7.5525 ms 7.5865 ms]                       
                        change: [-2.4643% -2.1081% -1.5816%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) high mild
  1 (1.00%) high severe

disasm_x86_detail       time:   [21.147 ms 21.200 ms 21.265 ms]                              
                        change: [-2.3077% -1.9077% -1.5143%] (p = 0.00 < 0.05)
                        Performance has improved.
```

PR #130, opened 2022-05-13
Add not_diet feature

clubby789:
This PR adds a new default feature, `not_diet` to both crates. If *disabled*, Capstone will be compiled in Diet mode, which would result in speed/size improvements. Currently, many tests will fail if run with this feature disabled, and will need to be gated behind it.

It might be more legible to invert this feature `#[cfg(feature = "diet")]` but I believe Cargo's dependency resolution [requires](https://www.reddit.com/r/rust/comments/sgegah/cargo_features_have_to_be_additive/) that features always _add_ new items, rather than removing them.

Rather than directly returning `None`, we could instead defer to Capstone's API (returning NULL pointers or empty strings) - this way is more explicit but with more repeated code.

### DIFF
--- a/capstone-rs/Cargo.toml
+++ b/capstone-rs/Cargo.toml
@@ -29,5 +29,8 @@ harness = false
 
 [features]
 default = ["full"]
+# The 'full' feature, enabled by default, compiles Capstone normally. When disabled,
+# Capstone will be built in Diet mode (https://www.capstone-engine.org/diet.html).
+# This disables some features to reduce the size of the library
 full = ["capstone-sys/full"]
 use_bindgen = ["capstone-sys/use_bindgen"]
--- a/capstone-rs/Cargo.toml
+++ b/capstone-rs/Cargo.toml
@@ -28,6 +28,6 @@ name = "my_benchmark"
 harness = false
 
 [features]
-default = ["not_diet"]
-not_diet = ["capstone-sys/not_diet"]
+default = ["full"]
+full = ["capstone-sys/full"]
 use_bindgen = ["capstone-sys/use_bindgen"]
--- a/capstone-rs/Cargo.toml
+++ b/capstone-rs/Cargo.toml
@@ -15,7 +15,7 @@ version = "0.11.0"
 travis-ci = { repository = "capstone-rust/capstone-rs" }
 
 [dependencies]
-capstone-sys = { path = "../capstone-sys", version = "0.15.0" }
+capstone-sys = { path = "../capstone-sys", version = "0.15.0",  default_features = false }
 libc = { version = "0.2", default-features = false }
 
 [dev-dependencies]
@@ -28,6 +28,6 @@ name = "my_benchmark"
 harness = false
 
 [features]
-default = []
-
+default = ["not_diet"]
+not_diet = ["capstone-sys/not_diet"]
 use_bindgen = ["capstone-sys/use_bindgen"]
--- a/capstone-rs/src/capstone.rs
+++ b/capstone-rs/src/capstone.rs
@@ -1,8 +1,8 @@
-use alloc::string::{String, ToString};
+use alloc::string::String;
 use core::convert::From;
 use core::marker::PhantomData;
 
-use libc::{c_int, c_uint, c_void};
+use libc::{c_int, c_void};
 
 use capstone_sys::cs_opt_value::*;
 use capstone_sys::*;
@@ -10,8 +10,10 @@ use capstone_sys::*;
 use crate::arch::CapstoneBuilder;
 use crate::constants::{Arch, Endian, ExtraMode, Mode, OptValue, Syntax};
 use crate::error::*;
-use crate::ffi::str_from_cstr_ptr;
 use crate::instruction::{Insn, InsnDetail, InsnGroupId, InsnId, Instructions, RegId};
+
+#[cfg(feature = "not_diet")]
+use {libc::c_uint, crate::ffi::str_from_cstr_ptr, alloc::string::ToString};
 
 /// An instance of the capstone disassembler
 ///
@@ -342,6 +344,8 @@ impl Capstone {
         result
     }
 
+
+    #[cfg(feature = "not_diet")]
     /// Converts a register id `reg_id` to a `String` containing the register name.
     pub fn reg_name(&self, reg_id: RegId) -> Option<String> {
         let reg_name = unsafe {
@@ -352,6 +356,13 @@ impl Capstone {
         Some(reg_name)
     }
 
+    #[cfg(not(feature = "not_diet"))]
+    /// `cs_reg_name` is not available in Diet mode.
+    pub fn reg_name(&self, _: RegId) -> Option<String> {
+        None
+    }
+
+    #[cfg(feature = "not_diet")]
     /// Converts an instruction id `insn_id` to a `String` containing the instruction name.
     ///
     /// Note: This function ignores the current syntax and uses the default syntax.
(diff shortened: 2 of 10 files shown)


clubby789:
Tests mostly pass now, but 3 issues:
```
---- test::test_arch_m68k_detail stdout ----
thread 'test::test_arch_m68k_detail' panicked at 'assertion failed: `(left == right)`
  left: `[M68kOperand(Reg(RegId(1))), M68kOperand(RegPair(RegId(5), RegId(6)))]`,
 right: `[]`: operands do not match for "0x1000: " (bytes=[4c, 00, 54, 04])', capstone-rs/src/test.rs:269:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- test::test_arch_ppc_detail stdout ----
thread 'test::test_arch_ppc_detail' panicked at 'assertion failed: `(left == right)`
  left: `[PpcOperand(Reg(RegId(88))), PpcOperand(Mem(PpcOpMem(ppc_op_mem { base: 0, disp: 0 })))]`,
 right: `[PpcOperand(Reg(RegId(88))), PpcOperand(Mem(PpcOpMem(ppc_op_mem { base: 10, disp: 0 })))]`: operands do not match for "0x1000: " (bytes=[80, 20, 00, 00])', capstone-rs/src/test.rs:269:5

---- test::test_arch_x86_detail stdout ----
thread 'test::test_arch_x86_detail' panicked at 'assertion failed: `(left == right)`
  left: `[X86Operand(X86Operand { size: 2, access: Some(WriteOnly), avx_bcast: X86_AVX_BCAST_INVALID, avx_zero_opmask: false, op_type: Reg(RegId(12)) }), X86Operand(X86Operand { size: 2, access: Some(ReadOnly), avx_bcast: X86_AVX_BCAST_INVALID, avx_zero_opmask: false, op_type: Mem(X86OpMem(x86_op_mem { segment: 0, base: 45, index: 0, scale: 1, disp: 50 })) })]`,
 right: `[X86Operand(X86Operand { size: 2, access: None, avx_bcast: X86_AVX_BCAST_INVALID, avx_zero_opmask: false, op_type: Reg(RegId(12)) }), X86Operand(X86Operand { size: 2, access: None, avx_bcast: X86_AVX_BCAST_INVALID, avx_zero_opmask: false, op_type: Mem(X86OpMem(x86_op_mem { segment: 0, base: 45, index: 0, scale: 1, disp: 50 })) })]`: operands do not match for "0x1000: " (bytes=[8d, 4c, 32])', capstone-rs/src/test.rs:269:5
```
Operands seem to being incorrectly processed in some arches

clubby789:
This behaviour seems to have been fixed in upstream - merging this PR on top of #129 makes all tests pass

capstone-rs/src/capstone.rs
```diff
@@ -342,6 +344,7 @@ impl Capstone {
         result
     }
 
+    #[cfg(feature = "not_diet")]
```
tmfink, line 347: I think we can avoid special casing this since the C library already checks if diet mode was enabled. I think we should avoid repeating logic in the C library where possible.

capstone-rs/src/test.rs
```diff
@@ -28,7 +26,9 @@ fn test_x86_simple() {
             Ok(insns) => {
                 assert_eq!(insns.len(), 2);
                 let is: Vec<_> = insns.iter().collect();
+                #[cfg(feature = "not_diet")]
```
tmfink, line 29: just put these 2 statements under a scope and put the cfg there

```rust
#[cfg(feature = "not_diet")]
{
    assert_eq!(is[0].mnemonic().unwrap(), "push");
    assert_eq!(is[1].mnemonic().unwrap(), "mov");
}
```

capstone-rs/src/test.rs
```diff
@@ -76,8 +78,10 @@ fn test_arm64_none() {
     assert!(cs.disasm_all(ARM_CODE, START_TEST_ADDR).unwrap().is_empty());
 }
 
+#[cfg(feature = "not_diet")]
 #[test]
 fn test_x86_names() {
+    use alloc::string::String;
```
tmfink, line 84: Let's put the import (protected by the cfg) at the top. We may need to use it in other function in the future.

capstone-rs/src/test.rs
```diff
@@ -180,15 +185,18 @@ fn test_detail_true() {
             let detail = cs
                 .insn_detail(&insns[insn_idx])
                 .expect("Unable to get detail");
-            let groups = detail.groups();
-            for insn_group_id in &insn_group_ids {
-                let insn_group = InsnGroupId(*insn_group_id as InsnGroupIdInt);
-                assert_eq!(groups.contains(&insn_group), false);
+            {
```
tmfink, line 188: is there a reason to add the new sub scope here?

capstone-rs/src/test.rs
```diff
@@ -272,6 +286,7 @@ fn test_instruction_group_helper<R: Copy + Into<RegId>>(
     expected_regs_write: &[R],
     has_default_syntax: bool,
 ) {
+    use std::collections::HashSet;
```
tmfink, line 289: hoist this import to the top (may be needed elsewhere in the future)

capstone-rs/src/test.rs
```diff
@@ -554,8 +572,8 @@ fn test_arch_mode_endian_insns(
     instructions_match(&mut cs_raw, expected_insns.as_slice(), true);
     instructions_match(&mut cs_raw_endian_set, expected_insns.as_slice(), true);
 }
-
-#[derive(Copy, Clone, Debug)]
+#[cfg_attr(feature = "not_diet", derive(Copy, Clone, Debug))]
```
tmfink, line 575: hoist common derives above

```rust
#[derive(Copy, Clone)]
...
```

capstone-sys/build.rs
```diff
@@ -133,16 +133,13 @@ fn build_capstone_cc() {
...
-        .flag_if_supported("-Wno-return-type")
-        .flag_if_supported("-Wno-implicit-fallthrough")
-        .flag_if_supported("-Wno-missing-field-initializers")
-        .flag_if_supported("-Wno-enum-conversion")
-        .static_crt(use_static_crt)
-        .compile("capstone");
+        // No need to display any warnings from the C library
+        .flag_if_supported("-w")
+        .static_crt(use_static_crt);
+    #[cfg(not(feature = "not_diet"))]
```
tmfink, line 139: put a newline before this line to make it visually separate

tmfink (changes requested):

tmfink:
Sorry for the late reply. Thanks for the changes--I've considered adding the diet support before.

capstone-rs/Cargo.toml
```diff
@@ -28,6 +28,6 @@ name = "my_benchmark"
 harness = false
 
 [features]
-default = []
-
+default = ["not_diet"]
+not_diet = ["capstone-sys/not_diet"]
```
tmfink, line 32: Can we think of a better name than `not_diet`? We need a "positive" term, but extra negatives are confusing.

Ideas:
- `full`
- `detail`

Also, please add a comment explaining what the feature does and include a link to the upstream documentation:
https://www.capstone-engine.org/diet.html

tmfink (commented):
Please add a similar feature to the [`cstool`](https://github.com/capstone-rust/capstone-rs/blob/c31409905a9fc4581133de248fcd0a66b2c37e36/cstool/Cargo.toml) crate in this workspace. It will be nice test of this feature.

capstone-rs/src/capstone.rs
clubby789: My thoughts behind organising it like this is it incurs unecessary runtime cost (calling into the library, loading the diet mode global) for a value we can determine statically

clubby789:
Any changes needed? The test failures look like a bug with Valgrind/the test harness itself

tmfink:
Thanks for the change and sorry for taking so long. I created #133 to track the Valgrind issue.

PR #131, opened 2022-06-22
Updated README.

IsaSca:
I was having difficulty getting the README example to compile. I had a look and I believe that this was because reg_names and group_names were expecting RegsID and IsnsGroupID respectively. 

I had a look and InsnDetail.regs_read(), and InsnDetail.regs_write() were returning RegsIter. While InsnDetail.groups() was returning  InsnGroupIter. So, I have updated the README to reflect this. If this is incorrect understanding of operation, please let me know. Thank you

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,19 +31,19 @@ See the [`capstone-sys`](capstone-sys) page for the requirements and supported p
 ```rust
 extern crate capstone;
 
-use capstone::prelude::*;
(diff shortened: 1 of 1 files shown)


tmfink:
Thanks for the PR! In order to avoid the README getting out of sync, I think I'll just have the README refer to an example program instead.

Issue #133, opened 2022-08-03
Valgrind failures in CI


tmfink:
Regular Valgrind tests fail in master branch. This could be a false positive.

https://github.com/capstone-rust/capstone-rs/runs/6753889756?check_suite_focus=true

clubby789:
Seems to be caused by https://stdrs.dev/nightly/x86_64-unknown-linux-gnu/src/std/sys/unix/fs.rs.html#157-160. The stdlib calls `statx` will a NULL pointer to check if the syscall is available. It seems to have been fixed in upstream Valgrind https://sourceware.org/git/?p=valgrind.git;a=commit;h=2a7d3ae768f9e5b29acd5cb743c3fb13640a391c

PR #134, opened 2022-09-07
ci: install valgrind from snap

tmfink:
Valgrind in Ubuntu repo shows false-positive against Rust std library.


tmfink:
Fixes #133

PR #143, opened 2023-08-06
[WIP] Compile-time Target Architecture Tags

### LINKED ISSUE #118: Proposal: add generic parameters to types to distinguish different data structures on different architectures
# Motivation

The link between core data types (such as `Capstone`, `Insn`, `InsnDetail`, etc.) and architecture-specific data types (the data types under `crate::arch`) is not that "explicit" on the typing. For example, to get the instruction ID on the x86 arch, we have to code:

```rust
let cs = Capstone::new().x86().build().unwrap();
let insns = cs.disasm_all(CODE, 0x1000).unwrap();
for i in insns.as_ref() {
    let insn_id = unsafe {
        // Neither typing nor documentation mention this!
        std::mem::transmute::<_, X86Insn>(i.id().0)
    };
    // do something with insn_id
}
```

which is not easier for beginners to catch.

We can simply resolve the problem above by adding a method `arch_insn_id` that returns the corresponding instruction ID enum variants just like the `InsnDetail::arch_detail` method:

```rust
pub enum ArchInsnId {
    X86InsnId(X86Insn),
    // other architectures
}

impl<'a> Insn<'a> {
    pub fn arch_insn_id(&self) -> ArchInsnId {
        // ... code
    }
}
```

This leads to another slightly disturbing problem. We have to `match` against the return value of `arch_insn_id` to extract the x86-specific instruction ID, given that **we are already confident about the architecture**. This problem also arises when we call the `InsnDetail::arch_detail` method or other methods with a similarly-typed return value.

# The Proposal

> The proposal posted here is only an (too-)early draft and more details may be missing for further considerations and discussions.

First of all, we can add a new trait that abstracts a specific architecture:

```rust
pub trait Arch {
    type InsnId;
    type InsnDetail;
    // ... other stuff
}
```

Then, we add a generic parameter to `Capstone`, `Insn` and `InsnDetail` that represents the architecture:

```rust
pub struct Capstone<A: Arch> {
    // ... fields
}

pub struct Insn<'a, A: Arch> {
    // ... fields
}

pub struct InsnDetail<'a, A: Arch> {
    // ... fields
}
```

Then, the methods mentioned in the motivation section can be typed in a more straight-forward way:

```rust
impl<A: Arch> Capstone<A> {
    pub fn insn_detail<'s, 'i: 's>(
        &'s self, 
        insn: &'i Insn<'_, A>
    ) -> CsResult<InsnDetail<'i, A>> {
        // ... code
    }
}

impl<'a, A: Arch> Insn<'a, A> {
    pub fn id(&self) -> A::InsnId {
        // ... code
    }
}

impl<'a, A: Arch> InsnDetail<'a, A> {
    pub fn arch_detail(&self) -> A::InsnDetail {
        // ... code
    }
}
```

No more `match`es, as long as we're targeting a specific architecture. Also, beginners can find corresponding architecture-specific implementations just by looking at the typings. The instruction ID problem can be resolved accordingly.

# Unresolved Problems

When the target architecture cannot be determined during compile-time (when the disassembler is created by the `Capstone::new_raw` method), the generic parameters cannot be set to represent specific architecture. To resolve this problem, maybe we need to introduce a `DynamicArch` that implements `Arch` and represents the target architecture is determined during runtime.

<hr />

This proposal may be pre-mature, but I do think that it reveals some (possibly minor) problems.

Lancern:
This PR resolves #118, which proposes **compile-time target architecture tags**. It contains a series of commits that implement the proposal. Problems may arise as I author these commits so I'm opening this PR in an early stage to make sure that we can agree on the specific designs in the end.

Here's a list of unresolved problems:
- [ ] Currently, there is no `CapstoneBuilder` that builds `Capstone` instances whose target architectures are "really" unknown at compile-time. When we call functions like `x86`, we have explicitly tell that the target architecture is `x86`. We should implement a new builder that sets the target architecture via a runtime `Arch` value.
- [ ] Currently, for many supported architectures, their `RegId` and `InsnGroupId` (some architectures even the `InsnId`) are simply defined as a bunch of `u32` values inside a mod rather than a dedicated enum (e.g. ARM, ARM64, x86, etc.). We’d better re-define them to be dedicated enums so that defining `ArchTag::RegId` and `ArchTag::InsnGroupId` is meaningful.
- [ ] After we support compile-time arch tags, many operations need not to return an `Option` or `Result` any more as they should always succeed given the exact target architecture. But for `DynamicArchTag` these operations may still fail so we need to keep their return types as it is now. How do we resolve the conflict?

### DIFF
--- a/capstone-rs/src/arch/arm.rs
+++ b/capstone-rs/src/arch/arm.rs
@@ -9,8 +9,10 @@ use capstone_sys::{
 use libc::c_uint;
 
 pub use crate::arch::arch_builder::arm::*;
-use crate::arch::DetailsArchInsn;
+use crate::arch::{ArchTag, DetailsArchInsn};
+use crate::arch::internal::ArchTagSealed;
 use crate::instruction::{RegId, RegIdInt};
+use crate::{Arch, InsnDetail};
 
 pub use capstone_sys::arm_insn_group as ArmInsnGroup;
 pub use capstone_sys::arm_insn as ArmInsn;
@@ -22,8 +24,36 @@ pub use capstone_sys::arm_cc as ArmCC;
 pub use capstone_sys::arm_mem_barrier as ArmMemBarrier;
 pub use capstone_sys::arm_setend_type as ArmSetendType;
 
+pub struct ArmArchTag;
+
+impl ArchTagSealed for ArmArchTag {}
+
+impl ArchTag for ArmArchTag {
+    type Builder = ArchCapstoneBuilder;
+
+    type Mode = ArchMode;
+    type ExtraMode = ArchExtraMode;
+    type Syntax = ArchSyntax;
+
+    type RegId = ArmReg::Type;
+    type InsnId = ArmInsn;
+    type InsnGroupId = ArmInsnGroup::Type;
+
+    type InsnDetail<'a> = ArmInsnDetail<'a>;
+
+    fn support_arch(arch: Arch) -> bool {
+        arch == Arch::ARM
+    }
+}
+
 /// Contains ARM-specific details for an instruction
 pub struct ArmInsnDetail<'a>(pub(crate) &'a cs_arm);
+
+impl<'a, 'i> From<&'i InsnDetail<'a, ArmArchTag>> for ArmInsnDetail<'a> {
+    fn from(value: &'i InsnDetail<'a, ArmArchTag>) -> Self {
+        Self(unsafe { &value.0.__bindgen_anon_1.arm })
+    }
+}
 
 /// ARM shift amount
 #[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
--- a/capstone-rs/src/arch/arm64.rs
+++ b/capstone-rs/src/arch/arm64.rs
@@ -1,13 +1,16 @@
 //! Contains arm64-specific types
 
+use core::convert::From;
+use core::{cmp, fmt, mem, slice};
+
+use capstone_sys::{arm64_op_mem, arm64_op_type, cs_arm64, cs_arm64_op};
 use libc::c_uint;
 
 pub use crate::arch::arch_builder::arm64::*;
-use crate::arch::DetailsArchInsn;
-use capstone_sys::{arm64_op_mem, arm64_op_type, cs_arm64, cs_arm64_op};
+use crate::arch::{ArchTag, DetailsArchInsn};
+use crate::arch::internal::ArchTagSealed;
 use crate::instruction::{RegId, RegIdInt};
-use core::convert::From;
-use core::{cmp, fmt, mem, slice};
+use crate::{Arch, InsnDetail};
 
 // Re-exports
 pub use capstone_sys::arm64_insn_group as Arm64InsnGroup;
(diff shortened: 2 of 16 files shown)


tmfink:
>  Currently, for many supported architectures, their `RegId` and `InsnGroupId` (some architectures even the `InsnId`) are simply defined as a bunch of `u32` values inside a mod rather than a dedicated enum (e.g. ARM, ARM64, x86, etc.). We’d better re-define them to be dedicated enums so that defining `ArchTag::RegId` and `ArchTag::InsnGroupId` is meaningful.

I like your change which uses [newtype_enum](https://docs.rs/bindgen/0.66.1/bindgen/struct.Builder.html#method.newtype_enum) instead of the [constified enum module](https://docs.rs/bindgen/0.66.1/bindgen/struct.Builder.html#method.constified_enum_module)! It's already a big improvement.

Bindgen didn't supported them when I first did this work. In fact, I contributed the "constified enum module" feature to bindgen (https://github.com/rust-lang/rust-bindgen/pull/741) to avoid the less ergonomic [constified enum](https://docs.rs/bindgen/0.66.1/bindgen/struct.Builder.html#method.constified_enum).

However, I think we may be able to do even better (although it might be more annoying). Ideally, we would expose actual Rust enums in the high-level bindings. That way, users could take advantage of the exhaustive pattern matching with `match`. The Capstone C code makes sure that 0 is always an "invalid" value such as `ARM_SYSREG_INVALID`. When converting from the C value to the Rust value, we could map any invalid value to an "invalid" variant of the enum OR always return an `Option<T>` when converting and return `None` if we get an invalid value.

```rust
// Have internal "invalid" variant
enum ArmReg {
    Invalid = 0,
    APSR,
    APSR_NZCV,
    // ...
}
```

```rust
// No invalid variant, so converting gives `Option<ArmReg>`
enum ArmReg {
    APSR = 1,
    APSR_NZCV,
    // ...
}
```

You don't need to implement this "Rust enum" change as a part of this PR--I just thought I'd let you know what I was thinking.

Lancern:
> Ideally, we would expose actual Rust enums in the high-level bindings.

Actually, I tried this at the beginning when writing the second commit. However, I quickly ran into problems.

Take ARM64 registers as an example. The problem is that there are actually 2 underlying C enums that both represent valid ARM64 reg IDs: [`arm64_reg`](https://github.com/capstone-rust/capstone-rs/blob/master/capstone-sys/capstone/include/capstone/arm64.h#L1010) and [`arm64_sysreg`](https://github.com/capstone-rust/capstone-rs/blob/master/capstone-sys/capstone/include/capstone/arm64.h#L1010). On the Rust side, each architecture can only have one Rust enum that represents an architecture register. Unfortunately, bindgen cannot merge two C enums into one big Rust enum. Also converting `Arm64Sysreg` into `Arm64Reg` won't work because the underlying value ranges of `Arm64Sysreg` and `Arm64Reg` are different. Thus the problem cannot be solved without considerable manual efforts if we generate Rust enums solely. 

On the other side, if we generate newtype enums for these two underlying C enums, we can easily convert an `Arm64Sysreg` into an `Arm64Reg` by manually implementing a `From`:

```rust
impl From<Arm64Sysreg> for Arm64Reg {
    fn from(sysreg: Arm64Sysreg): Self {
        Self(sysreg.0)
    }
}
```

tmfink:
> On the other side, if we generate newtype enums for these two underlying C enums, we can easily convert an `Arm64Sysreg` into an `Arm64Reg` by manually implementing a `From`:

Makes sense to me. If we want "handcrafted enums" that are more ergonomic, it would certainly be a separate PR.

Lancern:
I've updated some source code documentation to reflect the new architecture tag. However, I don't add a systematic overview or tutorial about the architecture tag as I'm not sure whether that's necessary. @tmfink Please have a look and see if the documentation can be further improved.

The examples and benches are already up-to-date with the latest API design. I'm checking whether the test coverage can be further improved and after that I belive this PR is ready to be formally reviewed and merged.

Lancern:
> Any plan to revive this?

Not yet, but I'm willing to. Are you working on this?

Lancern:
> > > Any plan to revive this?
> > 
> > 
> > Not yet, but I'm willing to. Are you working on this?
> 
> Not working on this, but ready to review.

Good to know that, maybe I'll try rebase this onto the latest main in the next few weeks.

Issue #84, opened 2020-05-06
regs_write() and regs_read() do not return complete results


bcantrill:
The instruction detail provided via `InsnDetail`'s `regs_write()` and `regs_read()` only return a subset of the registers written to or read from.

For example, running the C-based `cstool`:

```
$ cstool -d thumb "f0bd"
 0  f0 bd  pop	{r4, r5, r6, r7, pc}
	ID: 423 (pop)
	op_count: 5
		operands[0].type: REG = r4
		operands[0].access: WRITE
		operands[1].type: REG = r5
		operands[1].access: WRITE
		operands[2].type: REG = r6
		operands[2].access: WRITE
		operands[3].type: REG = r7
		operands[3].access: WRITE
		operands[4].type: REG = pc
		operands[4].access: WRITE
	Registers read: sp
	Registers modified: sp r4 r5 r6 r7 pc
	Groups: thumb thumb1only 
```

In contrast, running the capstone-rs-based `cstool`:

```
$ cstool -d --arch arm --mode thumb --code "f0bd"
      1000:  f0 bd                               pop     {r4, r5, r6, r7, pc}
             insn id:     423
             read regs:   sp
             write regs:  sp
             insn groups: thumb, thumb1only
```

The problem -- alluded to in #63 -- is that the C-based consumers are explicitly calling `cs_regs_access()` while the Rust-based consumers have no way of knowing that this must be called.  There are a number of ways of fixing this, but it seems that calling `insn_detail()` should return an `InsnDetail` for which `regs_write()`/`regs_read()` return correct data -- and therefore that `insn_detail()` should explicitly make the call to `cs_regs_access()`.  

Here is a diff that implements this:

```diff
diff --git a/capstone-rs/src/capstone.rs b/capstone-rs/src/capstone.rs
index 9e77f9d..1ae313d 100644
--- a/capstone-rs/src/capstone.rs
+++ b/capstone-rs/src/capstone.rs
@@ -374,6 +374,53 @@ impl Capstone {
         } else if Self::is_diet() {
             Err(Error::IrrelevantDataInDiet)
         } else {
+            /*
+             * To assure that our returned InsnDetail's regs_*() accessors
+             * return the correct data, we need to call into cs_regs_access().
+             * This interface is unsafe even by C's own poor standards for
+             * itself: it takes no bounds and has essentially undefined
+             * behavior if the arrays passed in correspond to the (same-sized)
[...]

PR #149, opened 2023-10-13
Regs access

joleeee:
I think the if cfg! is kinda weird and should probably rather have it just gate the entire function instead? Anyway, I'm not sure how to handle `assert!(len <= ints.len());` better, but it should never happen anyway so it probably doesn't matter?

Also I just picked 64 because python uses that, though I have a hard time imagining reading/writing to more than 64 registers: https://github.com/capstone-engine/capstone/blob/5fb8a423d4455cade99b12912142fd3a0c10d957/bindings/python/capstone/__init__.py#L929

### DIFF
--- a/capstone-rs/src/capstone.rs
+++ b/capstone-rs/src/capstone.rs
@@ -368,8 +368,23 @@ impl Capstone {
         }
     }
 
+    /// Get the registers are which are read to and written to
+    pub fn regs_access_buf(&self, insn: &Insn) -> CsResult<(Vec<RegId>, Vec<RegId>)> {
+        let mut read = Vec::new();
+        let mut write = Vec::new();
+
+        self.regs_access(insn, &mut read, &mut write)?;
+
+        Ok((read, write))
+    }
+
     /// Get the registers are which are read to and written to, in that order.
-    pub fn regs_access(&self, insn: &Insn) -> CsResult<(Vec<RegId>, Vec<RegId>)> {
+    pub fn regs_access(
+        &self,
+        insn: &Insn,
+        read: &mut Vec<RegId>,
+        write: &mut Vec<RegId>,
+    ) -> CsResult<()> {
         if cfg!(feature = "full") {
             Ok(unsafe {
                 let mut regs_read_count: u8 = 0;
@@ -391,15 +406,18 @@ impl Capstone {
                     return Err(err.into());
                 }
 
-                fn to_vec(ints: [u16; MAX_NUM_REGISTERS], len: usize) -> Vec<RegId> {
-                    assert!(len <= ints.len());
-                    ints[..len].iter().map(|v| RegId(*v)).collect()
-                }
-
-                (
-                    to_vec(regs_read, regs_read_count as usize),
-                    to_vec(regs_write, regs_write_count as usize),
-                )
+                read.extend(
+                    regs_read
+                        .iter()
+                        .take(regs_read_count as usize)
+                        .map(|x| RegId(*x)),
+                );
+                write.extend(
+                    regs_write
+                        .iter()
+                        .take(regs_write_count as usize)
+                        .map(|x| RegId(*x)),
+                );
             })
         } else {
             Err(Error::DetailOff)
--- a/capstone-rs/src/capstone.rs
+++ b/capstone-rs/src/capstone.rs
@@ -388,16 +388,15 @@ impl Capstone {
                     return Some(Err(err.into()));
                 }
 
-                assert!(regs_read_count as usize <= regs_read.len());
-                assert!(regs_write_count as usize <= regs_write.len());
+                fn to_vec(ints: [u16; 64], len: usize) -> Vec<RegId> {
+                    assert!(len <= ints.len());
+                    ints[..len].iter().map(|v| RegId(*v)).collect()
+                }
 
-                let regs_write: Vec<_> = regs_write[..regs_write_count as usize]
-                    .iter()
-                    .map(|v| RegId(*v))
-                    .collect();
-                let regs_read = regs_write[..regs_write_count as usize].to_vec();
-
-                (regs_write, regs_read)
+                (
+                    to_vec(regs_read, regs_read_count as usize),
+                    to_vec(regs_write, regs_write_count as usize),
+                )
             };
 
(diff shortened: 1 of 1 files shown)


capstone-rs/src/capstone.rs
```diff
@@ -365,6 +365,44 @@ impl Capstone {
         }
     }
 
+    /// Get the registers are which are read to and written to, in that order.
+    pub fn regs_access(&self, insn: &Insn) -> Option<CsResult<(Vec<RegId>, Vec<RegId>)>> {
```
tmfink, line 369: Return a `CsResult<...>` instead of an `Option<CsResult<...>>`.

You can return `Err(Error::DetailOff)` if the "full" feature is not enabled.

capstone-rs/src/capstone.rs
```diff
@@ -365,6 +365,44 @@ impl Capstone {
...
+                    assert!(len <= ints.len());
+                    ints[..len].iter().map(|v| RegId(*v)).collect()
+                }
+
+                (
+                    to_vec(regs_read, regs_read_count as usize),
+                    to_vec(regs_write, regs_write_count as usize),
+                )
+            }))
+        } else {
```
tmfink, line 401: use an early return in the case of an error to reduce the indentation of the "happy" case

tmfink:
> I think the if cfg! is kinda weird and should probably rather have it just gate the entire function instead? Anyway, I'm not sure how to handle `assert!(len <= ints.len());` better, but it should never happen anyway so it probably doesn't matter?

See the review comments (to come).

> Also I just picked 64 because python uses that, though I have a hard time imagining reading/writing to more than 64 registers: https://github.com/capstone-engine/capstone/blob/5fb8a423d4455cade99b12912142fd3a0c10d957/bindings/python/capstone/__init__.py#L929

64 is a fine value, but should probably be defined as a constant (magic numbers are not self-describing).

Looks like this would fix #63.

capstone-rs/src/capstone.rs
```diff
@@ -365,6 +365,44 @@ impl Capstone {
...
     }
 
+    /// Get the registers are which are read to and written to, in that order.
+    pub fn regs_access(&self, insn: &Insn) -> Option<CsResult<(Vec<RegId>, Vec<RegId>)>> {
+        if cfg!(feature = "full") {
+            Some(Ok(unsafe {
+                let mut regs_read_count: u8 = 0;
+                let mut regs_write_count: u8 = 0;
+
+                let mut regs_write = [0u16; 64];
```
tmfink, line 375: please pull the 64 constant to a constant with a name like `MAX_NUM_REGISTERS`. You can add a comment that the Python bindings use the same value.

capstone-rs/src/capstone.rs
tmfink, line 369: I don't like forcing allocations here by returning `Vec`. You can keep an allocation version for convenience, but please also create a `regs_access_buf()` where the user can provide their own buffers (which may just be stack allocated).

For simplicity, the `regs_access()` can call `regs_access_buf()`

tmfink, line 369: Instead of using a tuple, please make a new struct that makes it clear what is the read and write fields:

```rust
pub struct RegAccess {
    read: Vec<RegId>,
    write: Vec<RegId>,
}
```

The return type would be `CsResult<RegAccess>`.

capstone-rs/src/capstone.rs
```diff
@@ -365,6 +365,44 @@ impl Capstone {
...
+                    &mut regs_write as *mut _,
+                    &mut regs_write_count as *mut _,
+                );
+
+                if err != cs_err::CS_ERR_OK {
+                    return Some(Err(err.into()));
+                }
+
+                fn to_vec(ints: [u16; 64], len: usize) -> Vec<RegId> {
+                    assert!(len <= ints.len());
```
tmfink, line 392: the assert is not needed since the indexing below does a bounds check

tmfink (changes requested):

tmfink:
> Also @tmfink, whats the current status on the project? Seems like it's missing some updates? I could possibly step in and try and get things up to date.

I'll file a separate issue to address this question.

To update the internal capstone version, please:
1. Update the `update_capstone.rs` script and run it to update the C library:
    https://github.com/capstone-rust/capstone-rs/blob/7c9a51f6eb81709cde1de7f7498a0e16f29bd852/capstone-sys/scripts/update_capstone.sh#L6
1. [`capstone-sys/pre_generated/update-bindings.md`](https://github.com/capstone-rust/capstone-rs/blob/7c9a51f6eb81709cde1de7f7498a0e16f29bd852/capstone-sys/pre_generated/update-bindings.md)
1. Fix any build/test issues that come up.

tmfink:
> > Also @tmfink, whats the current status on the project? Seems like it's missing some updates? I could possibly step in and try and get things up to date.
> 
> I'll file a separate issue to address this question.

I just filed #151.
Thanks for asking about this. I've thought about this for a while but you finally gave me the push I needed. :smile:

capstone-rs/src/capstone.rs
joleeee: Should I just append the registers to the provided buffers, or should i clear them first?

capstone-rs/src/capstone.rs
joleeee: are you referring to the entire outer `if cfg!(...) {`? I agree it's cleaner, however the entire file is written this way currently :)

capstone-rs/src/capstone.rs
joleeee: If the user may provide stack allocated buffers, doesn't that mean they generally have to provide the length aswell? If they can pass a `regs_read: &mut [u16; MAX_NUM_REGISTERS]`, how will we specify how many were actually read?

joleeee: I used that struct but made the fields public, is it good now?

tmfink:
Addresses #84

tmfink:
Sorry for the long wait, but I merged the PR now.

I've decided to use this change as a starting point, but I'm going to make the regs access functionality more transparent as @bcantrill described in #84

Issue #148, opened 2023-10-10
Capstone v5 + documentation


joleeee:
What version of libcapstone does this use? Might be a good idea to document it aswell, as it's not clear to me if it's on 4 or 5.

joleeee:
on second thought it seems an old version of capstone was copied into this repo verbatim? https://github.com/capstone-rust/capstone-rs/tree/master/capstone-sys/capstone

why not use a git submodule?

tmfink:
As you mentioned, there is a copy of the Capstone C library in `capstone-sys/capstone/`. We used to have a git submodule, but I found it tedious/confusing, and [I'm not the only one](https://news.ycombinator.com/item?id=31792303). 

I also don't like how submodules refer to absolute URLs, so a given checkout is not self-contained. For example, if you clone this repo in your personal Git server instance, the submodule will still refer the GitHub Capstone repo.

For those reasons, I converted the submodule to a "flat checkout" in 87e315f98b624cdc824aae19cf31b7583e5f30bd.

I have an `update_capstone.sh` script to update the Capstone library version. The script defines what Git revision we pin to. I usually try pin to the latest revision in the `next` branch:

https://github.com/capstone-rust/capstone-rs/blob/7c9a51f6eb81709cde1de7f7498a0e16f29bd852/capstone-sys/scripts/update_capstone.sh#L6

joleeee:
Hmm... i patched this in on my own branch. I've never heard of any issues with git submodules really. I think it's much better than a copy paste from the capstone repos, getting rid of all commit logs from there etc, to me it seems to be bypassing the entire point of vcs, although of course we shouldn't do any patches there, but it's still strange to me.

It really confused me at first where the capstone source code was because i expected a submodule

> I also don't like how submodules refer to absolute URLs, so a given checkout is not self-contained. For example, if you clone this repo in your personal Git server instance, the submodule will still refer the GitHub Capstone repo.

I think that's sorta the point, but I can see why that can also be a drawback

joleeee:
I did get it working on my own branch, but i had to put a panic on some new ops i didn't know how to handle:

https://github.com/capstone-rust/capstone-rs/compare/master...joleeee:capstone-rs:v5

tmfink:
I closed this issue since I don't plan on changing the submodule and this issue seems to cover multiple things. Please file separate issues.

Also, please feel free to make a PR to update the Capstone version. Jumping major versions can take some work since the API may have changed in a backwards-incompatible way.

PR #150, opened 2023-10-13
use submodule instead of having capstone src in tree

joleeee:


tmfink:
Thanks for the contribution, but as mentioned in #148, I would like to avoid Git submodules. :sweat_smile:

PR #160, opened 2024-05-10
Fix cs_support enum bug

GarntS:
The enum `capstone::Arch` doesn't have the same values as capstone's `cs_arch` enum. This causes calls to `supports_arch()` to return incorrect results once the enum values diverge, which happens for riscv and a few others.

Fortunately, `capstone_sys::cs_arch` has a `From<capstone::Arch>` implementation, so we can just use that to do the correct conversion.


tmfink:
thanks for the fix!

PR #161, opened 2024-05-10
Implement InsnDetail (+ associated types) for sysz

GarntS:
This PR implements `InsnDetail`, `SysZOperand`, etc.  to enable instruction details for sysz. I used the RISC-V implementation as a template, so the code style should follow the rest of the library's conventions.

### DIFF
--- a/capstone-rs/src/arch/sysz.rs
+++ b/capstone-rs/src/arch/sysz.rs
@@ -1,3 +1,106 @@
 //! Contains sysz-specific types
 
+use core::convert::From;
+use core::{cmp, fmt, slice};
+
+// XXX todo(garnt): create rusty versions
+pub use capstone_sys::sysz_insn_group as SysZInsnGroup;
+pub use capstone_sys::sysz_insn as SysZInsn;
+pub use capstone_sys::sysz_reg as SysZReg;
+use capstone_sys::{cs_sysz, cs_sysz_op, sysz_op_mem, sysz_op_type};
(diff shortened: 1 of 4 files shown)


capstone-rs/src/instruction.rs
```diff
@@ -413,6 +413,7 @@ impl<'a> InsnDetail<'a> {
                             $detail($insn_detail(unsafe { &self.0.__bindgen_anon_1.$arch }))
                         }
                     )*
+                    #[allow(unreachable_patterns)]
```
tmfink, line 416: you can just delete this catch-all pattern arm

capstone-rs/src/arch/sysz.rs
```diff
@@ -1,3 +1,106 @@
 //! Contains sysz-specific types
```
tmfink, line 1: Please add a `test_arch_sysz_detail()` in `capstone-rs/src/test.rs` (you can see similar tests for other archs).

tmfink (changes requested):
thanks for the change!

capstone-rs/src/arch/sysz.rs
GarntS: Added this test

GarntS:
Hey - made the changes you requested. Not 100% sure how Github's PRs work.

tmfink:
Thanks for the contribution!

Issue #97, opened 2021-04-29
ArmOperand "access" member of type Option<RegAccessType> is missing


flipnut:
Capstone's C types `cs_arm_op` and `cs_x86_op` etc. have a member `uint8_t access`, that provides info about the operand being read or written to by the instruction. In the rust world, the related struct `X86Operand` has the member `pub access: Option<RegAccessType>,`. Howevery, for `ArmOperand` the member is missing.

Unfortunately, I am not familiar with the semi-automated translation process of c types to rust types in capstone-rust. With some guidance I am happy to help where I can to get this fixed.

enjhnsn2:
Has there been any progress on this issue?

tmfink:
Sorry for the late reply. The Rust types are "bespoke" and written by hand. :sweat_smile: 

If you are interested in contributing the feature, take a look at `RegAccessType` and the use for X86:

- `RegAccessType` in `capstone-rs/src/instruction.rs`:
https://github.com/capstone-rust/capstone-rs/blob/cb4cc5e7e3e3b0d1e88265a650a920392b8d5ab5/capstone-rs/src/instruction.rs#L42
- `impl TryFrom<cs_ac_type> for RegAccessType`: https://github.com/capstone-rust/capstone-rs/blob/cb4cc5e7e3e3b0d1e88265a650a920392b8d5ab5/capstone-rs/src/instruction.rs#L71
- X86 implementation for `RegAccessType` https://github.com/capstone-rust/capstone-rs/blob/cb4cc5e7e3e3b0d1e88265a650a920392b8d5ab5/capstone-rs/src/arch/x86.rs#L52

You will also need to update the tests in `capstone-rs/src/test.rs`.

Let me know if you have any questions.

PR #163, opened 2024-08-15
Make RegAccessType available for arm.

daniel-dt-thoma:
fixes capstone-rust/capstone-rs#97

I've copied the approach from x86 to arm as you've suggested and adjusted the tests accordingly.

### DIFF
--- a/capstone-rs/src/arch/arm.rs
+++ b/capstone-rs/src/arch/arm.rs
@@ -1,16 +1,16 @@
 //! Contains arm-specific types
 
-use core::convert::From;
+use core::convert::{From, TryInto};
 use core::{cmp, fmt, slice};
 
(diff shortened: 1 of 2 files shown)


capstone-rs/src/arch/arm.rs
```diff
@@ -129,6 +129,10 @@ pub struct ArmOperand {
 
     /// Operand type
     pub op_type: ArmOperandType,
+
+    /// How is this operand accessed? NOTE: this field is irrelevant if engine
+    /// is compiled in DIET mode.
```
tmfink, line 134: Instead, lets say that the field is always `None` in DIET mode.

tmfink:
Thanks for your contribution!

PR #166, opened 2025-01-14
Fix segfault in cs_option() on s390x

iii-i:
capstone-rs crashes on s390x (also knows as SystemZ in the LLVM world)
as follows:

    Thread 2 received signal SIGSEGV, Segmentation fault.
    0x000002aa00a6a604 in cs_option (ud=140702726002672, type=CS_OPT_SYNTAX, value=2) at capstone/cs.c:782
    782             return arch_configs[handle->arch].arch_option(handle, type, value);

    (gdb) p/x handle->arch
    $7 = 0x7ffb

handle->arch is clearly corrupted. It is assigned as follows in
cs_open():

       0x000002aa00a6a44a <+146>:   lgr     %r9,%r3
    [...]
       0x000002aa00a6a468 <+176>:   rosbg   %r9,%r10,0,31,32
    [...]
    458                     ud->arch = arch;
    459                     ud->mode = mode;
       0x000002aa00a6a47a <+194>:   stg     %r9,0(%r13)

Here the 32-bit arch value and the 32-bit mode value are placed into a
64-bit register and stored with one instruction.

The important part is that cs_mode is a 32-bit value passed in %r3, and
the code uses instructions with mnemonics ending with "g", which
operate on full 64-bit registers [1]. This is allowed because of the
s390x ABI [1]: it requires zero-extension of 4-byte int arguments.

Bitfield enums are currently generated like this:

    #[repr(C)]
    [...]
    pub struct cs_mode(pub libc::c_uint);

which causes the problem above, because the ABI does not require
zero-extension of 4-byte struct arguments.

Starting from RustTarget::Stable_1_28, bindgen
generates #[repr(transparent)] instead of #[repr(C)], which is
designed to solve exactly this kind of problems [3].

capstone-rs' MSRV is 1.70, so it should be possible to go even further,
but be conservative and use the minimum version necessary to fix the
issue.

[1] https://publibfp.dhe.ibm.com/epubs/pdf/a227832d.pdf
[2] https://github.com/IBM/s390x-abi/releases
[3] https://doc.rust-lang.org/nomicon/other-reprs.html#reprtransparent

### DIFF
--- a/capstone-rs/src/constants.rs
+++ b/capstone-rs/src/constants.rs
@@ -232,69 +232,69 @@ define_cs_enum_wrapper!(
         => Mode = cs_mode
     ]
     /// 32-bit ARM
-    => Arm = CS_MODE_ARM;
+    => Arm = { cs_mode::CS_MODE_ARM };
     /// 16-bit mode (X86)
-    => Mode16 = CS_MODE_16;
+    => Mode16 = { cs_mode::CS_MODE_16 };
     /// 32-bit mode (X86)
-    => Mode32 = CS_MODE_32;
+    => Mode32 = { cs_mode::CS_MODE_32 };
     /// 64-bit mode (X86, PPC)
-    => Mode64 = CS_MODE_64;
+    => Mode64 = { cs_mode::CS_MODE_64 };
     /// ARM's Thumb mode, including Thumb-2
-    => Thumb = CS_MODE_THUMB;
+    => Thumb = { cs_mode::CS_MODE_THUMB };
     /// Mips II ISA
-    => Mips2 = CS_MODE_MIPS2;
+    => Mips2 = { cs_mode::CS_MODE_MIPS2 };
     /// Mips III ISA
-    => Mips3 = CS_MODE_MIPS3;
+    => Mips3 = { cs_mode::CS_MODE_MIPS3 };
     /// Mips32r6 ISA
-    => Mips32R6 = CS_MODE_MIPS32R6;
+    => Mips32R6 = { cs_mode::CS_MODE_MIPS32R6 };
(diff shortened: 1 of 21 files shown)


iii-i:
* Fixed a couple typos.

philmd (approved):

iii-i:
* (Hopefully) fix a clippy warning in unrelated code.

iii-i:
* Local testing with nightly has produced even more clippy warnings; fixed all of them.

tmfink:
Thanks for the PR! Your comments are very descriptive.
I'll try to review your changes, but no promises on the timeline.

capstone-rs/src/constants.rs
```diff
@@ -232,69 +232,69 @@ define_cs_enum_wrapper!(
         => Mode = cs_mode
     ]
     /// 32-bit ARM
-    => Arm = CS_MODE_ARM;
+    => Arm = { cs_mode::CS_MODE_ARM };
```
tmfink, line 235: to avoid the need to write `cs_mode::` many times, can we instead modify the `define_cs_enum_wrapper!` macro to include the enum automatically?

tmfink (changes requested):

capstone-rs/src/constants.rs
iii-i: The problem is that, e.g., `CS_OPT_SYNTAX_INTEL`, which is also used with this macro, is not generated as an enum member:

```
pub mod cs_opt_value {
[...]
    pub const CS_OPT_SYNTAX_INTEL: Type = 1;
```

We could probably indicate to the macro that this is the case, but I thought the extra complexity would not be worth it.

Issue #169, opened 2025-03-21
Capstone 6.0.0 alpha support


terahlunah:
Hi, I tried to update `capstone-rs` to support `capstone` 6.0.0 alpha 3, but I was not familiar enough with capstone and didn't succeed.

Do you have any plan to start working on an `alpha` branch of `capstone-rs` to support `capstone` 6.0.0 alphas (and the final release eventually)?

Thanks for your work!

tmfink:
> Hi, I tried to update `capstone-rs` to support `capstone` 6.0.0 alpha 3, but I was not familiar enough with capstone and didn't succeed.

I'm happy to mentor with doing updates.
Update the `CAPSTONE_REVISION` variable in [capstone-sys/scripts/update_capstone.sh](https://github.com/capstone-rust/capstone-rs/blob/478d7655daad7d4d0e6ad5bed7d1042ee33d5016/capstone-sys/scripts/update_capstone.sh) and run the script to update the copy of the Capstone C library.
Then, you need to fixup `capstone-sys`/`capstone-rs` to handle the changes to the C library.

> Do you have any plan to start working on an `alpha` branch of `capstone-rs` to support `capstone` 6.0.0 alphas (and the final release eventually)?

I have no immediate plans, but as mentioned, I happy to mentor anyone who wants to try the update.

> Thanks for your work!

Your welcome 😄

jiegec:
Capstone 6.0.0 is an exciting upgrade, but I am unsure whether we should bump before the official release. When will the 6.0.0 official release come out?

terahlunah:
> Then, you need to fixup capstone-sys/capstone-rs to handle the changes to the C library.

Updating `capstone-sys` was fairly straightforward, but I failed to adapt `capstone-rs` to the new code.

> I am happy to maintain an alpha branch though.

That would be greatly appreciated!

jiegec:
> > Then, you need to fixup capstone-sys/capstone-rs to handle the changes to the C library.
> 
> Updating `capstone-sys` was fairly straightforward, but I failed to adapt `capstone-rs` to the new code.
> 
> > I am happy to maintain an alpha branch though.
> 
> That would be greatly appreciated!

I'm working on it. Stay tuned!

terahlunah:
Thank you so much @jiegec! 🚀

PR #172, opened 2025-05-08
rs, sys: bump capstone to 5.0.6

jiegec:
- Bump capstone to 5.0.6 using update_capstone.sh
- Regenerated Rust bindings using bindgen
- Handle new op types: SVCR and SME index
- Add TriCore support from capstone 5.0.6

Tested via `cargo test`.


jiegec:
Will try to upgrade to capstone 6.0.0 later (#169)

PR #174, opened 2025-05-14
Make `build.rs` compatible with Bazel's sandbox

AlexDenisov:
When building with Bazel's sandbox, all the capstone source files are actually symlinks. In this case `build.rs` doesn't find any files to build and attempts to link the static library right away.

Reproducer and the error messages can be found [here](https://github.com/bazelbuild/rules_rust/issues/3435).
I tested the change locally on macOS with the following spec:
```bazel
crate.spec(
    package = "capstone-sys",
    git = "https://github.com/AlexDenisov/capstone-rs.git",
    branch = "bazel-compatibility",
)
```
which is working as expected.

Is there anything else I'd need to check to ensure this change doesn't break anything?

### DIFF
--- a/capstone-sys/build.rs
+++ b/capstone-sys/build.rs
@@ -91,7 +91,7 @@ fn build_capstone_cc() {
                 .file_type()
                 .expect("Failed to read capstone source directory");
             let file_name = e.file_name().into_string().expect("Invalid filename");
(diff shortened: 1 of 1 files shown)


tmfink:
Thanks for the fix!

PR #188, opened 2025-07-29
add check_only feature to disable building the C code

RalfJung:
I am looking into ways for how we can speed up check-builds in Miri (see https://github.com/rust-lang/miri/issues/4461). The libffi build script accounts for a significant fraction of the remaining build time here since the C code still gets built, even when that is entirely unnecessary for a check build.

In the rustc workspace, we have a `check_only` feature on some crates for exactly this purpose. I wonder if you'd be open to the idea of having a similar feature for libffi? Of course I would prefer to actually have native support in cargo for this but that idea has a bunch of unanswered design questions (https://github.com/rust-lang/cargo/issues/4001). So meanwhile, the best we can do is hack it in ourselves. This is done with a cargo feature to have cargo properly cache the build with and without the feature and avoid rebuilds when the flag changes. I'm happy to bikeshed the name, e.g. making it more clear that this is an unstable/unsupported configuration if you prefer that.  Could you imagine landing something like this?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,8 +37,8 @@ jobs:
       - name: test
         run: env ${{ matrix.rust.env }} ./capstone-rs/ci/test.sh
 
-      - name: test (all features)
-        run: env ${{ matrix.rust.env }} ALL_FEATURES=1 ./capstone-rs/ci/test.sh
+      - name: test (with bindgen)
+        run: env ${{ matrix.rust.env }} FEATURES="use_bindgen" ./capstone-rs/ci/test.sh
 
       - name: test (only enable x86 and arm64)
         run: env ${{ matrix.rust.env }} FEATURES=std,full,arch_x86,arch_arm64 NO_DEFAULT_FEATURES=1 ./capstone-rs/ci/test.sh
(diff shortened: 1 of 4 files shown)


RalfJung:
This feature enables the mode of skipping the C builds...

If this becomes negative then I don't know how to use it from Miri.

RalfJung:
In particular, if this becomes negative and Miri depends on another crate that depends on capstone, it will be impossible for Miri to choose check-only mode.

So I agree features should generally be positive, but this is not a normal feature.

jiegec:
That's a hard decision...

RalfJung:
I would think about it like this:
why do we say "features are additive"? Because if crate A enables a feature and crate B does not, then the feature ends up being enabled, so if a feature removes a function everything would break.
This feature isn't about adding/removing functions that would change whether the code compilers. Even using a feature at all for this is kind of a hack. So the usual heuristic does not immediately apply. The question is more, if crate A requests a check-only build, and crate B does not, what should happen? I would argue that the result should be a check-only build -- if that breaks something (i.e. if someone enables this feature this on a `cargo build`, not just a `cargo check`), then whoever did that clearly did something wrong.

jiegec:
@tmfink What's your opinion on this? I am glad to apply any changes required.

PR #189, opened 2025-07-30
rs, sys: rename and invert check_only feature to build

jiegec:
See [discussions](https://github.com/capstone-rust/capstone-rs/pull/188#issuecomment-3135421713).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,8 +37,8 @@ jobs:
       - name: test
         run: env ${{ matrix.rust.env }} ./capstone-rs/ci/test.sh
 
-      - name: test (with bindgen)
-        run: env ${{ matrix.rust.env }} FEATURES="use_bindgen" ./capstone-rs/ci/test.sh
+      - name: test (all features)
+        run: env ${{ matrix.rust.env }} ALL_FEATURES=1 ./capstone-rs/ci/test.sh
 
       - name: test (only enable x86 and arm64)
-        run: env ${{ matrix.rust.env }} FEATURES=std,full,arch_x86,arch_arm64 NO_DEFAULT_FEATURES=1 ./capstone-rs/ci/test.sh
+        run: env ${{ matrix.rust.env }} FEATURES=std,full,arch_x86,arch_arm64,build NO_DEFAULT_FEATURES=1 ./capstone-rs/ci/test.sh
--- a/capstone-rs/Cargo.toml
+++ b/capstone-rs/Cargo.toml
@@ -29,7 +29,7 @@ name = "my_benchmark"
 harness = false
 
 [features]
-default = ["full", "std", "support_all_archs"]
+default = ["full", "std", "support_all_archs", "build"]
 std = []
 # The 'full' feature, enabled by default, compiles Capstone normally. When disabled,
 # Capstone will be built in Diet mode (https://www.capstone-engine.org/diet.html).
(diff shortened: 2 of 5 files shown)


jiegec:
`cargo clean && time cargo check`: 7.56s; `cargo clean && time cargo check --no-default-features --features full,std,support_all_archs`: 1.41s.

tmfink:
After thinking about our discussion w/ @RalfJung in #188 some more, I think there are issues if we have the "check only" behavior controlled by a feature:

## Option 1: `check_only` feature that is not enabled by default

If any dependency depends on capstone-rs w/ the `--all-features`, then `check_only` will be used and `cargo build` will not work.

## Option 2: use inverted `build` feature that is enabled by default

This PR (#189) uses this approach. We now have the opposite issue.
If any dependency that depends on capstone doesn't use `default-features = false`, then capstone-rs crate will always build the Capstone C library.
That means there's essentially no point in the feature (other than doing builds of the crate directly).

## Proposal: use environment variable

Features are meant to be used when you want a change in functionality/API and are enabled if any dependency in your dependency graph depends on a feature.
However, only the final application (not any library dependency) should determine whether or not we are using "check only" behavior.

Users who want more efficient check/miri runs could set a variable like `CAPSTONE_CHECK_ONLY=1`:

```sh
# Example in sh-like shell
CAPSTONE_CHECK_ONLY=1 cargo check
```

> [!NOTE]  
> I'm not saying `CAPSTONE_CHECK_ONLY` is the best name, I'm just using it as an example

This environment variable approach is used by crates like
* [openssl](https://docs.rs/openssl/0.10.73/openssl/index.html#manual): `OPENSSL_DIR`, `OPENSSL_NO_VENDOR`, ...
* [clang_sys](https://github.com/KyleMayes/clang-sys/blob/1a6210e34bfa8ba9098575831518dbe2523f77e1/README.md#environment-variables): `LLVM_CONFIG_PATH`, `LIBCLANG_PATH`, ...

Only the final application should decide if capstone uses the "check only" behavior.

@RalfJung, @jiegec thoughts?

jiegec:
I can't find a perfect solution. If it doesn't work, I can only revert everything and wait for cargo upstream to come up with a better solution.

RalfJung:
I don't think I understand the problem with "option 1". I think there are no visible problems for crates depending on capstone. Is it that for capstone development itself, `--all-features` doesn't work any more?

tmfink:
> I don't think that's possible? A cargo manifest can have `default-features = false`, but there is no `all-features = true`.

@RalfJung You're right--there's not "Cargo.toml" equivalent of `cargo build --all-features`.

> > Proposal: use environment variable
> 
> I don't think this works. You'll have to add this env var to `rerun-if-env-changed` and then each time one switches between `cargo build` and `CAPSTONE_CHECK_ONLY=1 cargo check`, it will re-build the C dependency.

You're right--I was thinking about using `rerun-if-env-changed` but that will cause unnecessary rebuilds of the C library.

@jiegec if you agree, could you revert this PR (#189) to go back to the original "check_only" behavior that @RalfJung had?
This will make local development of capstone slightly more annoying/complicated since `--all-features` won't work, but it will (hopefully) make our users more happy and shouldn't break them.

Hopefully https://github.com/rust-lang/cargo/issues/4001 will be resolved so we can remove this hack. :smiling_face_with_tear:

PR #197, opened 2026-02-28
feat: Add a method to Arm64InsnDetail to expose the post_index flag

PandasAreBears:
The `post_index` flag was added to `cs_arm64` in this commit: https://github.com/capstone-rust/capstone-rs/commit/f8ae2293753b899e6b6340b41a4c62cbcb0f3243

This change exposes the value of the flag in `Arm64InsnDetail`.

Thanks

### DIFF
--- a/capstone-rs/src/test.rs
+++ b/capstone-rs/src/test.rs
@@ -1186,6 +1186,49 @@ fn test_arch_arm_detail() {
     );
 }
 
+#[cfg(feature = "arch_arm")]
+#[test]
+fn test_arch_arm_writeback() {
+    let mut cs = Capstone::new()
+        .arm()
+        .mode(arm::ArchMode::Arm)
+        .build()
+        .unwrap();
+    cs.set_detail(true).unwrap();
+
+    // ldr r0, [r1]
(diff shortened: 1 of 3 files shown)


jiegec (approved):

tmfink (changes requested):
I like the change, but could you add the same change to `capstone-rs/src/arch/arm.rs`?
It would also be great if you could add a unit test for the new methods.

PandasAreBears:
I've added a test which asserts the correctness of the `writeback` and `post_index` fields. It wasn't clear how to cleanly slot it into the existing `test_arch_mode_endian_insns` pattern. There probably a nicer way of writing this test but I don't have any more time today.

capstone-rs/src/test.rs
```diff
@@ -1186,6 +1186,49 @@ fn test_arch_arm_detail() {
     );
 }
 
+#[cfg(feature = "arch_arm64")]
+#[test]
+fn test_arch_arm64_writeback() {
```
jiegec, line 1234: Maybe another test for arm?

jiegec (approved):

PR #198, opened 2026-05-29
sys: only compile enabled arch sources in build.rs

CodeWenjiu:
In GCC 14+, -Wimplicit-function-declaration became an error that -w flag cannot suppress. The auto-generated AArch64 sources in capstone v6 trigger
this error even when the AArch64 feature is disabled, because build.rs previously compiled all arch directories unconditionally.

Filter arch directories by enabled cargo features, so only selected architectures' sources are compiled.

### DIFF
--- a/capstone-sys/build.rs
+++ b/capstone-sys/build.rs
@@ -121,7 +121,7 @@ fn build_capstone_cc() {
             "x86" => cfg!(feature = "arch_x86"),
             "xcore" => cfg!(feature = "arch_xcore"),
             "xtensa" => cfg!(feature = "arch_xtensa"),
-            _ => panic!("Unhandled arch dir {dir_name}"),
+            _ => panic!("Unhandled arch dir {}", dir_name),
         }
     }
 
--- a/capstone-sys/build.rs
+++ b/capstone-sys/build.rs
@@ -96,7 +96,6 @@ fn build_capstone_cc() {
     }
 
     fn is_arch_enabled(dir_name: &str) -> bool {
-        #[allow(clippy::match_like_matches_macro)]
         match dir_name.to_lowercase().as_str() {
             "aarch64" => cfg!(feature = "arch_aarch64"),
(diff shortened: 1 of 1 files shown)


tmfink:
@CodeWenjiu thanks for the enhancement! I thought we we already doing to this in 335ca9ec26257e6c5c2deecf29a8286faf81f9fe, but I guess we were not. :laughing:

Could you fix the `clippy::match_like_matches_macro` warning in the code? It looks like this may be a false-positive from clippy, so you may just want to add the `#[allow(clippy::match_like_matches_macro)]` attribute on that match expression.

capstone-sys/build.rs
```diff
@@ -95,12 +95,42 @@ fn build_capstone_cc() {
...
+            "sh" => cfg!(feature = "arch_sh"),
+            "sparc" => cfg!(feature = "arch_sparc"),
+            "systemz" => cfg!(feature = "arch_systemz"),
+            "tms320c64x" => cfg!(feature = "arch_tms320c64x"),
+            "tricore" => cfg!(feature = "arch_tricore"),
+            "wasm" => cfg!(feature = "arch_wasm"),
+            "x86" => cfg!(feature = "arch_x86"),
+            "xcore" => cfg!(feature = "arch_xcore"),
+            "xtensa" => cfg!(feature = "arch_xtensa"),
+            _ => false,
```
tmfink: Let's actually make the `_` case a panic. That way, we need to explicitly handle every case. For example, if we don't support a platform, we will can explicitly handle that with a `false` match case.

```rust
match {
    /* ... */
    _ => panic!("Unhandled arch dir {dir_name}"),
}
```

CodeWenjiu: Changed `_` arm to `panic!()` and dropped the `#[allow]` since the panic makes it unnecessary.

jiegec (approved):

tmfink (approved):

PR #201, opened 2026-06-26
Fix unused lifetime in ArchDetail (#199) and add CI feature combination tests (#200)

### LINKED ISSUE #199: Build fails with all features diabled
I tried both the release version and the wip "capstone-next" PR and both branches gave me the following error:
```
error[E0392]: lifetime parameter `'a` is never used
   --> /home/theo/.cargo/git/checkouts/capstone-rs-6f9cee1e7a49c9fe/e2c56d3/capstone-rs/src/arch/mod.rs:780:29
    |
780 |         pub enum ArchDetail<'a> {
    |                             ^^ unused lifetime parameter
...
910 | detail_arch_base!(detail_defs);
    | ------------------------------ in this macro invocation
    |
    = help: consider removing `'a`, referring to it in a field, or using a marker such as `PhantomData`
    = note: this error originates in the macro `detail_defs` whic
```

### LINKED ISSUE #200: CI should test more feature combinations
To avoid issues like #199, we should test more combinations of features.
For simplicity, we can use [cargo-hack](https://github.com/taiki-e/cargo-hack), which supports testing multiple features.

There may be too many features to do a full powerset set, but we should at least test:
- all features disabled
- all features enabled
- each feature individually enabled (no others)
- random subset of some features enabled (e.g. 2 architectures enabled, 3 architectures enabled)

jiegec:
Fix #199: Add `_Lifetime(PhantomData<&'a ()>)` variant to `ArchDetail` enum to ensure the `'a` lifetime parameter is always used, even when no arch features are enabled. Also add `#[non_exhaustive]` to `ArchDetail` and `ArchOperand` since their variants depend on feature flags.

Fix #200: Add `feature-checks` CI job using a Python script that dynamically discovers `arch_*` features from `Cargo.toml` and tests:
- all features disabled
- all features enabled
- each arch feature individually
- random subsets (2-5 arch features, shuffled per trial)

### DIFF
--- a/scripts/feature-checks.py
+++ b/scripts/feature-checks.py
@@ -1,0 +1,95 @@
+#!/usr/bin/env python3
+"""Test capstone-rs with various feature combinations.
+
+Tests:
+  - all features disabled
+  - all features enabled
+  - each arch feature individually enabled
+  - random subsets (2-arch and 3-arch combinations)
+"""
+
+import random
+import re
+import subprocess
+import sys
+import os
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+MANIFEST = REPO_ROOT / "capstone-rs" / "Cargo.toml"
+CARGO = os.environ.get("CARGO", "cargo")
+
+failed = []
+
+
+def cargo_check(label, *args):
+    cmd = [CARGO, "check", "--manifest-path", str(MANIFEST)] + list(args)
+    print(f"\n=== {label} ===")
+    print(f"    {' '.join(cmd)}")
+    result = subprocess.run(cmd, capture_output=True, text=True)
+    if result.returncode == 0:
+        print("    PASS")
+    else:
+        print(result.stdout)
+        print(result.stderr)
+        print("    FAIL")
+        failed.append(label)
+
+
+def get_arch_features():
+    """Extract arch feature names from Cargo.toml."""
+    text = MANIFEST.read_text()
+    features = []
+    for line in text.splitlines():
+        m = re.match(r"^(arch_\w+) = \[\"capstone-sys/arch_", line)
+        if m:
+            features.append(m.group(1))
+    return sorted(features)
+
+
+def main():
+    arch_features = get_arch_features()
+    print(f"Found {len(arch_features)} arch features:")
+    for f in arch_features:
+        print(f"  {f}")
+
+    # 1. All features disabled
+    cargo_check("no-default-features", "--no-default-features")
+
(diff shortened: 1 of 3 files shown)


scripts/feature-checks.py
```diff
@@ -0,0 +1,95 @@
...
+import subprocess
+import sys
+import os
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+MANIFEST = REPO_ROOT / "capstone-rs" / "Cargo.toml"
+CARGO = os.environ.get("CARGO", "cargo")
+
+failed = []
```
tmfink: If you use a global (especially one you modify), then make it ALL CAPS so that it's clearly a global.

However, I suggest you avoid the mutable global and instead have `cargo_check()` return an error (or append to a `failed` argument) so it's more clear how everything is tracked. Sometimes we need a global, but I don't think we do here.

scripts/feature-checks.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def cargo_check(label, *args):
```
tmfink: Instead of requiring cargo args, I think we should instead exposing "pythonic" args:
- `default_features=False`
- `all_features=False`
- `features=None`: list of features

scripts/feature-checks.py
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env python3
```
tmfink, line 1: For convenience, you can make this file executable (`chmod +x`)--git tracks these file permissions.

scripts/feature-checks.py
tmfink: I suggest adding python type hints to make the script easier to understand. If you use an LSP like basedpyright, it helps catch warnings/errors. Linters like ruff can also help.

https://docs.python.org/3/library/typing.html

scripts/feature-checks.py
```diff
@@ -0,0 +1,95 @@
...
+        print(f"  {f}")
+
+    # 1. All features disabled
+    cargo_check("no-default-features", "--no-default-features")
+
+    # 2. All features enabled
+    cargo_check("all-features", "--all-features")
+
+    # 3. Each arch feature individually
+    base = "std,full"
```
tmfink: let's also have tests for different combinations of std/full being enabled/disabled.

tmfink (changes requested):
This change looks really good--just some comments on the python script

tmfink (approved):
Looks good

